Issue #4, opened 2016-02-17
Remove OCHamcrest imports from Unit and Functional Tests Project.


tirodkar:
Since we have a wrapper around OCHamcrest, we do not need an explicit import nor any linking against it in the Test Projects themselves.

Remove all instances of `-f OCHamcrest` from the `UnitTests.xcodeproj` and `FunctionalTest.xcodeproj`. Also, remove any uneeded imports containing import `<OCHamcrest/...>`.

Issue #6, opened 2016-02-17
Missing script


bzarzoza:
build_universal_static_framework.sh is missing

tirodkar:
Brandon, does this break your project? It's because an extra file is being added in the Scripts/ group of EarlGrey.xcodeproj. You could remove this Group and the error should be gone.

bzarzoza:
Yes, it does break (non-cocoapod) integration.

tirodkar:
@bzarzoza, we removed the script references from the project, could you confirm if you still see the issue?

bzarzoza:
Looks good. It would be cool to see the script, too.

tirodkar:
It was just a simple script that we were using before embedded framework were supported. Closing this for now. Do tell us if you require any more info.

PR #7, opened 2016-02-17
Remove Extra Build Script Reference and extra flag from Tests.

tirodkar:
An extra file - `https://github.com/google/EarlGrey/pull/7` is present in the `EarlGrey.xcodeproj` file, which isn't used and causes project breaks.

Also remove `-framework OCHamcrest` from both Test Projects as per https://github.com/google/EarlGrey/issues/4. Also, since we have a number of non-modular imports in UnitTests, turn `Enable Modules` off to prevent issues with imports.


khandpur:
LGTM.

tirodkar:
This fixes the file issue in https://github.com/google/EarlGrey/issues/6.

PR #14, opened 2016-02-20
Fixing travis build

axi0mX:
clean makes it build, and the scheme is not configured for building, so it should just be test

tested locally after a git clone and it worked


axi0mX:
I signed it.

axi0mX:
I signed it.

khandpur:
Thank you for this fix!

PR #28, opened 2016-02-24
Fix document about importing library in Swift

dreamagicjp:
To import EarlGrey library in Swift, `#import <EarlGrey/EarlGrey.h>` should be written in test bundle's bridging header.


tirodkar:
Thanks for this, @dreamagicjp

PR #66, opened 2016-04-03
Experimental Changes

axi0mX:
Experimental changes.


bootstraponline:
> c) other apps, such as Safari

Does that work on real devices as well?

axi0mX:
@bootstraponline Yes! You can control the entire device, for now. 

Allowing control outside of the app's sandbox is a potential security issue, so it's possible Apple will block it in the future, but for now it works, I tested it on iOS 9.3.

bootstraponline:
The PR was closed? :(

khandpur:
We discussed a better design offline with the author hence decided to put this on hold.

PR #78, opened 2016-04-29
Add badges for license/cocoapods/travis ci

bootstraponline:
[ci skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,7 @@
 # EarlGrey
+[![Apache License](https://img.shields.io/badge/license-Apache%202-lightgrey.svg?style=flat)](https://github.com/google/EarlGrey/blob/master/LICENSE)
+[![CocoaPods](https://img.shields.io/cocoapods/v/EarlGrey.svg?maxAge=2592000)](https://cocoapods.org/pods/EarlGrey)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,8 @@
 # EarlGrey
+[![Apache License](https://img.shields.io/badge/license-Apache%202-lightgrey.svg?style=flat)](https://github.com/google/EarlGrey/blob/master/LICENSE)
+[![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
```
khandpur: Do we filly support cartage without your pending PR which adds cartage spec?

bootstraponline: That's a good point. Carthage is fully supported but only if you know how to manually configure the testing target which is a complex and painful process. I'll remove the Carthage badge for now.

bootstraponline:
Updated.

README.md
tirodkar: Yup. We should add a Carthage badge only when we provide a "Carthage Installation Steps" guide in the docs.

khandpur:
Thanks for adding these badges.

PR #85, opened 2016-05-16
Cleaning up error messages and removing unused states from the state.

khandpur:
- Clean up error messages
- Remove unused statetracker states
- fix for a bug that applied search action one last time just before a timeout without checking the hierarchy again. As a result, the screenshot was different from when the actual failure was expected to occur.

### DIFF
--- a/EarlGrey/Core/GREYElementInteraction.m
+++ b/EarlGrey/Core/GREYElementInteraction.m
@@ -97,12 +97,13 @@ NSString *const kGREYAssertionErrorUserInfoKey = @"kGREYAssertionErrorUserInfoKe
  *          seconds, a timeout error will be produced and no elements will be returned.
  */
 - (NSArray *)matchedElementsWithTimeout:(NSTimeInterval)timeout error:(__strong NSError **)error {
+  NSParameterAssert(error);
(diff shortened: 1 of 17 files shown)


.gitignore
```diff
@@ -2,5 +2,3 @@
 *.sw[o-p]
 .DS_Store
 xcuserdata
-*.gem
```
khandpur: will revert this.

EarlGrey/Core/GREYElementInteraction.m
```diff
@@ -97,12 +97,13 @@ - (instancetype)inRoot:(id<GREYMatcher>)rootMatcher {
  *          seconds, a timeout error will be produced and no elements will be returned.
  */
 - (NSArray *)matchedElementsWithTimeout:(NSTimeInterval)timeout error:(__strong NSError **)error {
+  NSParameterAssert(error);
```
sid-github: nit: also assert timeout param: i.e. `timeout > 0` as NSTimeInterval is double and a negative value can be passed in.

sid-github:
Waiting for travis but LG aside from a nit.

PR #88, opened 2016-05-17
Run Ruby gem tests on Travis

bootstraponline:


bootstraponline:
This is working now! Ready for review.

.travis.yml
```diff
@@ -20,10 +21,17 @@ script:
   - set -o pipefail
   - xcodebuild -version
   - xcodebuild -showsdks
+  - if [[ ${GREY_RUBY} ]]; then
+      cd gem;
+      bundle install --retry=3;
```
tirodkar: Do all travis machines come with rake and bundler installed?

bootstraponline: No. The one we're using does.

.travis.yml
```diff
@@ -20,10 +21,17 @@ script:
...
+      rake;
+    fi
   - if [[ ${DESTINATION} == *"OS=9.2"* || ${DESTINATION} == *"OS=8.4"* ]]; then
       env NSUnbufferedIO=YES xcodebuild -project Tests/UnitTests/UnitTests.xcodeproj -scheme EarlGreyUnitTests -sdk "$SDK" -destination "$DESTINATION" -configuration Debug ONLY_ACTIVE_ARCH=NO test | xcpretty -c;
     fi
   - if [[ ${DESTINATION} == *"OS=9.3"* ]]; then
       env NSUnbufferedIO=YES xcodebuild -project Tests/FunctionalTests/FunctionalTests.xcodeproj -scheme EarlGreyFunctionalSwiftTests -sdk "$SDK" -destination "$DESTINATION" -configuration Debug ONLY_ACTIVE_ARCH=NO test | xcpretty -c;
     fi
-  - env NSUnbufferedIO=YES xcodebuild -project Tests/FunctionalTests/FunctionalTests.xcodeproj -scheme EarlGreyFunctionalTests -sdk "$SDK" -destination "$DESTINATION" -configuration Debug ONLY_ACTIVE_ARCH=NO test | xcpretty -c;
+  - if [[ ${DESTINATION} ]]; then
```
khandpur: Destination should always be defined by Travis...no?

.travis.yml
khandpur: Do we know what why `bundle install` would require retries? Is it common that git & network requests would fail?

.travis.yml
khandpur: Ah. Never mind. You're adding a new matrix.

.travis.yml
bootstraponline: Yes, it's common.

bootstraponline: Rails, a well known Ruby project, [uses the same setting.](https://github.com/rails/rails/blob/d6e7334d19c234d42904e4a70b5a76831731b114/.travis.yml#L18)

PR #90, opened 2016-05-17
remove NSAssert and add object when not nil

fernyb:
This change fixes my tests.

### DIFF
--- a/EarlGrey/Provider/GREYElementProvider.m
+++ b/EarlGrey/Provider/GREYElementProvider.m
@@ -149,7 +149,8 @@
               // Replace mock views with the views they encapsulate.
               element = [element view];
             }
-            if(element) {
+            // if elemnet is not nil add to runningElementHierarchy, sometimes element can be nil
+            if (element) {
               [runningElementHierarchy addObject:element];
             }
           }
--- a/EarlGrey/Provider/GREYElementProvider.m
+++ b/EarlGrey/Provider/GREYElementProvider.m
@@ -149,8 +149,9 @@
               // Replace mock views with the views they encapsulate.
               element = [element view];
             }
-            NSAssert(element, @"element should not be nil");
-            [runningElementHierarchy addObject:element];
+            if(element) {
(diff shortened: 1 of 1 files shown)


fernyb:
I signed it!

EarlGrey/Provider/GREYElementProvider.m
```diff
@@ -149,8 +149,9 @@ - (NSEnumerator *)dataEnumerator {
               // Replace mock views with the views they encapsulate.
               element = [element view];
             }
-            NSAssert(element, @"element should not be nil");
-            [runningElementHierarchy addObject:element];
+            if(element) {
+              [runningElementHierarchy addObject:element];
+            }
```
tirodkar: Could you also add a comment as to why we're doing this?
Also, you could have the same check in GREYElementHierarchy.m:137 - 

      // Need to add a nil check since sometimes accessibilityElement is nil.
      if (accessibilityElement) {
        [subViewSet addObject:accessibilityElement];
      }

EarlGrey/Provider/GREYElementProvider.m
```diff
@@ -149,8 +149,9 @@ - (NSEnumerator *)dataEnumerator {
               // Replace mock views with the views they encapsulate.
               element = [element view];
             }
-            NSAssert(element, @"element should not be nil");
-            [runningElementHierarchy addObject:element];
+            if(element) {
```
khandpur: nit: space after "if"

EarlGrey/Provider/GREYElementProvider.m
```diff
@@ -149,8 +149,10 @@ - (NSEnumerator *)dataEnumerator {
               // Replace mock views with the views they encapsulate.
               element = [element view];
             }
-            NSAssert(element, @"element should not be nil");
-            [runningElementHierarchy addObject:element];
+            // if elemnet is not nil add to runningElementHierarchy, sometimes element can be nil
```
tirodkar: element*

PR #93, opened 2016-05-18
In case of an iOS 8.1 type action, make sure we check if autocorrect is supported on the keyboard delegate.

tirodkar:


khandpur:
@tirodkar As per our offline discussions - is this ready for review yet?

tirodkar:
This should be good to go.

PR #101, opened 2016-05-27
Add note about animations

bootstraponline:
> https://pspdfkit.com/blog/2016/running-ui-tests-with-ludicrous-speed/

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -114,3 +114,25 @@ For the Example project, run the `EarlGreyExampleSwiftTests` target once then fi
 For physical device builds, replace `Debug-iphonesimulator` with `Debug-iphoneos`.
 
 You can verify the `EarlGrey.framework` is included in the bundle at `EarlGreyExampleSwiftTests.xctest/Frameworks/EarlGrey.framework`
+
+**How should I handle animations?**
+
+Infinite animations must be disabled as they'll cause problems with EarlGrey tests.
+
+```swift
+public class UIAutomation {
+    public static let animate:Bool = Process.arguments.indexOf("NO_ANIMATE") == nil
+}
+
+if UIAutomation.animate { ... }
+```
+
+For finite animations, the animation speed can be increased. This allows UIKit
+completion blocks and async calls to execute as they normally would, just faster. This matches the
+real conditions the iOS app is run under and will catch more bugs than simply disabling animations.
+Note that the speedup doesn't work on `UIScrollView` because it animates via `CADisplayLink` internally.
+Refer to the [PSPDFKit blog post for more details.](https://pspdfkit.com/blog/2016/running-ui-tests-with-ludicrous-speed/)
+
+```swift
+UIApplication.sharedApplication().keyWindow?.layer.speed = 100
+```
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -117,18 +117,22 @@ You can verify the `EarlGrey.framework` is included in the bundle at `EarlGreyEx
 
 **How should I handle animations?**
 
-Infinite animations must be disabled as they'll cause problems with EarlGrey tests.
+By default, [EarlGrey truncates CALayer based animations](../EarlGrey/Common/GREYConfiguration.h#L108) that exceed a threshold. The max animation duration setting is configurable:
 
 ```swift
-public class UIAutomation {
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -114,3 +114,18 @@ For the Example project, run the `EarlGreyExampleSwiftTests` target once then fi
...
+
+```swift
+UIView.AnimationsEnabled = false
+```
+
+Alternatively, the animation speed can be increased to improve coverage of the app.
+This enables finding bugs that otherwise wouldn't be triggered if we simply disabled animations.
+
+```swift
+UIApplication.sharedApplication().keyWindow!.layer.speed = 100
```
albinekcom: The force unwrapping is not needed so it can be written in this style:
`UIApplication.sharedApplication().keyWindow?.layer.speed = 100`

docs/faq.md
```diff
@@ -114,3 +114,18 @@ For the Example project, run the `EarlGreyExampleSwiftTests` target once then fi
 For physical device builds, replace `Debug-iphonesimulator` with `Debug-iphoneos`.
 
 You can verify the `EarlGrey.framework` is included in the bundle at `EarlGreyExampleSwiftTests.xctest/Frameworks/EarlGrey.framework`
+
+**How should I handle animations?**
+
+Disabling animations improves test performance and eliminates a source of flakiness.
+
+```swift
+UIView.AnimationsEnabled = false
```
albinekcom: It should be:
`UIView.setAnimationsEnabled(false)`

bootstraponline: Fixed

docs/faq.md
bootstraponline: Fixed

docs/faq.md
```diff
@@ -114,3 +114,18 @@ For the Example project, run the `EarlGreyExampleSwiftTests` target once then fi
 For physical device builds, replace `Debug-iphonesimulator` with `Debug-iphoneos`.
 
 You can verify the `EarlGrey.framework` is included in the bundle at `EarlGreyExampleSwiftTests.xctest/Frameworks/EarlGrey.framework`
+
+**How should I handle animations?**
+
+Disabling animations improves test performance and eliminates a source of flakiness.
```
khandpur: As per our chat with [Peter](https://github.com/steipete), shouldn't we advice speeding up animations as opposed to disabling it completely so that delegate callbacks and completion blocks are executed in an orderly fashion?

bootstraponline: The same argument can be made for Android and I don't see anyone advocating speeding up animations there. Instead the advice, consistently, has been to [disable animations](https://youtu.be/isihPOY2vS4?t=1302). That's true both for system animations and app specific animations. To the extent EarlGrey is Espresso for iOS, it'd be nice to not have a completely opposite position.

I'll update the FAQ so the pros/cons are more obvious, specifically the trade off between more coverage via animations and the resulting flakiness.

bootstraponline:
Updated.

bootstraponline:
Updated with new language that recommends disabling only infinite animations and using the `layer.speed` trick for everything else.

docs/faq.md
```diff
@@ -114,3 +114,25 @@ For the Example project, run the `EarlGreyExampleSwiftTests` target once then fi
 For physical device builds, replace `Debug-iphonesimulator` with `Debug-iphoneos`.
 
 You can verify the `EarlGrey.framework` is included in the bundle at `EarlGreyExampleSwiftTests.xctest/Frameworks/EarlGrey.framework`
+
+**How should I handle animations?**
+
+Infinite animations must be disabled as they'll cause problems with EarlGrey tests.
```
khandpur: Forgot to mention:
EarlGrey has a way to force cyclic animations to finish sooner using kGREYConfigKeyCALayerMaxAnimationDuration config key.

```
[[GREYConfiguration sharedInstance] setValue:@(kMaxAnimationInterval)
                                forConfigKey:kGREYConfigKeyCALayerMaxAnimationDuration];
```

bootstraponline:
Updated:
 - removed process args
 - added kGREYConfigKeyCALayerMaxAnimationDuration

bootstraponline:
Updated again.

khandpur:
LGTM

Issue #103, opened 2016-05-31
Reset application state before each EarlGrey test?


bootstraponline:
What's the recommend approach for resetting the application state before each test? I'm thinking of issues such as:
- logging out
- dismissing any app specific dialogs

- [discussion group topic](https://groups.google.com/forum/#!topic/earlgrey-discuss/8JpItL8e7kk)

bootstraponline:
If we had XCUITest support then we could just relaunch the app.

khandpur:
FWIW, we're still working on XCUITest support design and plan on sharing that doc on Slack for feedback.

I don't quite agree that XCUITest alone will solve all these problems (more details below).

For the recommended approach to reset the app, you'd need to:
1) Logout:
This is useful if you want to test different login flows but if you just care about getting past the sign-in screen, structure the tests in a way so that if the login screen is shown, it will try to login otherwise assert that the main screen is shown. Alternatively, a better approach would be to fake out the login flow (for tests that don't care about login scenario). Keep in mind that keychain entries will be persisted so even with XCUITest you'd have to somehow make sure it's faked out or works with App that's already signed in.
2) Dismiss dialogs
This should be done as part of the tearDown. You can tell if a dialog is shown by inspecting the ui hierarchy.
3) Make sure all the persistent state is cleared so app starts in a clean state: This would involve some work. The best way would be to bake a resetApplication method in the app delegate which resets services, databases, tears down windows and recreates them in a clean state. Even XCUITest will suffer from this because app has persistent data which survives subsequent launches.

bootstraponline:
> bake a resetApplication method in the app delegate which resets services, databases, tears down windows and recreates them in a clean state.

That makes sense. Thanks for the explanation.

khandpur:
Any other updates on this? Is it safe to close?

bootstraponline:
Closed.

PR #107, opened 2016-06-01
Add resetApplication to FAQ

bootstraponline:
#103

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -150,3 +150,45 @@ This is difficult to do when an element is duplicated (same label/class/location
 
 We recommend combining the matchers [as suggested here](api.md#earlgrey-matchers) and then adding
 `grey_interactable()` or `grey_sufficientlyVisible()`.
+
+**How do I reset application state before each test?**
+
+In the application target's Build Settings, set **Defines Module** to **Yes**. Create a `resetApplicationForTesting()` method on the AppDelegate. The reset method will be invoked on `setUp` instead of `tearDown` because otherwise there's no guarentee the app will be in a clean state when the first test is run.
+
+Swift:
+In the EarlGrey test target, import the application using `@testable`. In `setUp()`, aquire a reference to the delegate then invoke `resetApplicationForTesting()`.
+
+```swift
+// swift
+@testable import App
+
+class MyTests: XCTestCase {
+
+    override func setUp() {
+        super.setUp()
+
+        let delegate: App.AppDelegate = (UIApplication.sharedApplication().delegate as? App.AppDelegate)!
+        delegate.resetApplicationForTesting()
+    }
+```
+
+
+Objective C:
+In the EarlGrey test target, import the application's app delegate header. In `setUp()` aquire a pointer to the delegate then invoke `resetApplicationForTesting()`.
+
+```objc
+// Objective C
+#import "MyAppDelegate.h"
+
+@interface MyTests : XCTestCase
+@end
+
+@implementation MyTests
+
+- (void)setUp {
+    [super setUp];
+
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -142,3 +142,23 @@ Refer to the [PSPDFKit blog post for more details.](https://pspdfkit.com/blog/20
 ```swift
 UIApplication.sharedApplication().keyWindow?.layer.speed = 100
 ```
+
+**How do I reset application state before each test using Swift?**
+
+In the application's Build Settings, set **Defines Module** to **Yes**. Create a `resetApplication()` method on the AppDelegate.
```
khandpur: application target's Build Settings?

docs/faq.md
```diff
@@ -142,3 +142,23 @@ Refer to the [PSPDFKit blog post for more details.](https://pspdfkit.com/blog/20
 ```swift
 UIApplication.sharedApplication().keyWindow?.layer.speed = 100
 ```
+
+**How do I reset application state before each test using Swift?**
```
khandpur: how about in objc?

docs/faq.md
```diff
@@ -142,3 +142,23 @@ Refer to the [PSPDFKit blog post for more details.](https://pspdfkit.com/blog/20
...
+
+In the EarlGrey test target, import the application using `@testable`. In `setUp()`, aquire a reference to the delegate then invoke `resetApplication()`.
+
+```swift
+// swift
+@testable import App
+
+class MyTests: XCTestCase {
+
+    override func setUp() {
```
khandpur: wouldn't it be more intuitive to reset in tearDown?

bootstraponline: No because we want to ensure reset runs before each test run. If we had a previous run without reset, then we're already in a bad state and resetting in tearDown doesn't help us.

docs/faq.md

docs/faq.md
bootstraponline: Added.

docs/faq.md
khandpur: In what scenarios would the previous run not clean up the state? Are you worried about the app crashing or the same device previously having the app installed?

In a clean simulator scenario, when a test is run for the first time, shouldn't the app be in a clean state? So then, in setUp method you would:
1. assert the state is clean
2. setup the required mocks and fake data
3. Create other test fixtures for the app.

Then test case begins, modifies the state of the app and in tearDown, the app state is cleaned again, mocks are removed, fake data is reset, etc. And the cycle repeats...

bootstraponline: > Are you worried about the app crashing or the same device previously having the app installed?
> In a clean simulator scenario, when a test is run for the first time, shouldn't the app be in a clean state?

I'm worried about the app already being on the simulator / used for manual testing. We have no guarantees about the "clean simulator scenario" when doing EarlGrey testing.

khandpur: SG. I would mention that to be the reason why we recommend doing reset in setUp

bootstraponline: Updated.

docs/faq.md
```diff
@@ -150,3 +150,45 @@ This is difficult to do when an element is duplicated (same label/class/location
 
 We recommend combining the matchers [as suggested here](api.md#earlgrey-matchers) and then adding
 `grey_interactable()` or `grey_sufficientlyVisible()`.
+
+**How do I reset application state before each test?**
+
+In the application targets's Build Settings, set **Defines Module** to **Yes**. Create a `resetApplicationForTesting()` method on the AppDelegate. The reset method will be invoked on `setUp` instead of `tearDown` because otherwise there's no guarentee the app will be in a clean state when we run our first test.
```
khandpur: small nit:
targets's => target's

bootstraponline: Fixed

PR #112, opened 2016-06-03
Document that grey_allOf order matters

bootstraponline:


docs/api.md
```diff
@@ -118,6 +118,10 @@ id<GREYMatcher> visibleSendButtonMatcher =
     performAction:grey_tap()];
 ```
 
+Note that with `grey_allOf` the order matters. If `grey_sufficientlyVisible` is used first, then every element
+in the entire application will be checked for visibility. It's important to order matchers from
+most selective (such as label and id) to least.
```
khandpur: label and id => accessibility label and accessibility id

docs/faq.md
```diff
@@ -192,3 +192,18 @@ In the EarlGrey test target, import the application's app delegate header. In `s
     [delegate resetApplicationForTesting];
 }
 ```
+
+**How do I increase the speed of `grey_allOf`?**
```
khandpur: Shouldn't the question be more along the lines of what you initially reported:
Takes too long to do XXX or screen appears frozen for X seconds?

docs/faq.md
```diff
@@ -192,3 +192,18 @@ In the EarlGrey test target, import the application's app delegate header. In `s
...
+
+Make sure the matchers are ordered from most specific to least. For example:
+
+`grey_allOf(grey_accessibilityID(@"Foo"), grey_sufficientlyVisible()`
```
khandpur: grey_allOf should to be nil terminated.
grey_allOf(foo, bar, nil)

docs/faq.md
```diff
@@ -192,3 +192,18 @@ In the EarlGrey test target, import the application's app delegate header. In `s
...
+
+will find one element with the target id and then check that single element for visibility.
+If we had the order wrong:
+
+`grey_allOf(grey_sufficientlyVisible(), grey_accessibilityID(@"Foo")`
```
khandpur: nil terminate

docs/faq.md
bootstraponline: is that true for swift as well? we've been using it like this:

```swift
EarlGrey().selectElementWithMatcher(grey_allOfMatchers(
            grey_kindOfClass(NSClassFromString("_UIAlertControllerView")),
            grey_interactable()))

```

khandpur: you're right, it's just for objc

bootstraponline:
Updated.

docs/faq.md
```diff
@@ -192,3 +192,41 @@ In the EarlGrey test target, import the application's app delegate header. In `s
     [delegate resetApplicationForTesting];
 }
 ```
+
+**Why does the screen appear frozen for 30 seconds?**
```
khandpur: ironically 30 seconds is also the timeout for interaction so if there's an ongoing background animation or a synchronization bug that renders the app busy for ~ 30 seconds then it may appear as through screen is frozen but really it's just waiting for idle. Though, in such cases, EarlGrey will fail with a timeout instead.

bootstraponline: good to know, updated with this info.

khandpur:
Thanks for this!

PR #113, opened 2016-06-03
Add NSClassFromString tip

bootstraponline:


docs/faq.md
```diff
@@ -192,3 +192,12 @@ In the EarlGrey test target, import the application's app delegate header. In `s
     [delegate resetApplicationForTesting];
 }
 ```
+
+**How do I match on internal classes?**
```
khandpur: Perhaps something like:
How do I create a matcher that matches internal UIKit classes?

is a bit more descriptive?

docs/faq.md
```diff
@@ -192,3 +192,12 @@ In the EarlGrey test target, import the application's app delegate header. In `s
...
+
+Use `NSClassFromString` to match on internal classes (prefixed with `_`) that
```
khandpur: There are UIKit classes that are not prefixed with _ but still considered internal / private because they're not directly accessible.

docs/faq.md
```diff
@@ -192,3 +192,12 @@ In the EarlGrey test target, import the application's app delegate header. In `s
...
+can't be referenced directly.
+
+```swift
+grey_kindOfClass(NSClassFromString("_UIAlertControllerView")
```
khandpur: missing closing parenthesis at the end.

docs/faq.md
bootstraponline: agreed

docs/faq.md
bootstraponline: cool, I didn't know. updated.

docs/faq.md
bootstraponline: fixed

PR #116, opened 2016-06-03
Document EarlGrey detection in FAQ

bootstraponline:
I had been using a launch argument for a similar reason. The `NSClassFromString` detection method is significantly more reliable.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -153,10 +153,10 @@ We recommend combining the matchers [as suggested here](api.md#earlgrey-matchers
 
 **How do I reset application state before each test?**
 
-In the application target's Build Settings, set **Defines Module** to **Yes**. Create a `resetApplicationForTesting()` method on the AppDelegate. The reset method will be invoked on `setUp` instead of `tearDown` because otherwise there's no guarentee the app will be in a clean state when the first test is run.
+In the application target's Build Settings, set **Defines Module** to **Yes**. Create a `resetApplicationForTesting()` method on the AppDelegate. The reset method will be invoked on `setUp` instead of `tearDown` because otherwise there's no guarantee the app will be in a clean state when the first test is run.
 
 Swift:
-In the EarlGrey test target, import the application using `@testable`. In `setUp()`, aquire a reference to the delegate then invoke `resetApplicationForTesting()`.
+In the EarlGrey test target, import the application using `@testable`. In `setUp()`, acquire a reference to the delegate then invoke `resetApplicationForTesting()`.
 
 ```swift
 // swift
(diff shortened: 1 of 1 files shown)


khandpur:
I don't think that's more robust though. It can easily break if we refactor and rename EarlGreyImpl to something else. A better check would be via env vars that's should be made part of the scheme that's running those tests.

bootstraponline:
> It can easily break if we refactor and rename EarlGreyImpl to something else.

If `EarlGreyImpl` is ever going to rename then I agree, this isn't a great approach.

I don't know how to set env vars or process args on test clouds like Amazon, all we get to send is an xctest bundle. The `NSClassFromString` approach is my preferred method for that reason. 

What do you think about defining an empty class that's guaranteed not to change? I've seen some tests check for `XCTest` but that doesn't differentiate between unit and UI targets.

khandpur:
ah...sigh. that really sucks. Worth asking them how to set env vars so you can just do:
#ifdef EARLGREY_TESTENV
...<earlgrey specific code>...
#endif

khandpur:
Spoke on slack about it. Let's move forward with mentioning both ways. #ifdef and runtime check for presence of class because in some environments it's not possible to check if code is running on earlgrey or not.

bootstraponline:
I think we want to use a [build configuration](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/InteractingWithCAPIs.html#//apple_ref/doc/uid/TP40014216-CH8-ID34) (instead of ENV variables) because the `#ifdef ...` syntax doesn't appear to work with environment variables. 

> Build configurations include the literal true and false values, command line flags, and the platform and language-version testing functions listed in the table below. You can specify command line flags using -D <#flag#>.
> 
> https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/InteractingWithCAPIs.html#//apple_ref/doc/uid/TP40014216-CH8-ID34

bootstraponline:
Updated to demo both build config & runtime check.

khandpur:
Oops, you're right. Nice catch - I meant build settings.

docs/faq.md
```diff
@@ -241,3 +241,22 @@ grey_allOf(grey_sufficientlyVisible(),
 then all elements in the entire application will be checked for visibility, and finally one
 with a matching id will be selected. It's significantly faster to use the most targeted
 matchers first (typically `grey_accessibilityID` or `grey_accessibilityLabel`).
+
+**How can I detect if I'm running in an EarlGrey target?**
+
+Creating a [build configuration](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/InteractingWithCAPIs.html#//apple_ref/doc/uid/TP40014216-CH8-ID34) for EarlGrey will allow conditional compliation:
```
khandpur: s.p.
compliation => compilation

docs/faq.md
```diff
@@ -241,3 +241,22 @@ grey_allOf(grey_sufficientlyVisible(),
...
+
+```swift
+#if EARLGREY_ENV
+...
+#else
+...
+#endif
```
khandpur: any reason why endif isn't highlighted just like #if and #else?

docs/faq.md

khandpur:
waiting for conflicts to be resolved before merging.

bootstraponline:
Fixed!

PR #122, opened 2016-06-07
Add Swift conditional example

### LINKED ISSUE #118: Swift conditional example
Is this the correct way to handle pointers with Swift?

I think this would be useful as a FAQ entry. It works in my test suite.

```swift
GREYCondition(name: "Wait for collection view to populate", block: { _ in
    let errorOrNil = UnsafeMutablePointer<NSError?>.alloc(1)
    errorOrNil.initialize(nil)

    EarlGrey().selectElementWithMatcher(collectionViewMatcher)
              .assertWithMatcher(grey_notNil(), error: errorOrNil)

    let success = errorOrNil.memory == nil
    errorOrNil.destroy()
    errorOrNil.dealloc(1)

    return success
}).waitWithTimeout(20.0)
```

bootstraponline:
Fix #118

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -292,3 +292,23 @@ EarlGrey.selectElementWithMatcher(matcher)
                onElementWithMatcher:grey_kindOfClass([UITableView class])
                   assertWithMatcher:grey_notNil()];
 ```
+
+**How do I wait for an element to appear using Swift?**
+
+The best way is to [setup synchronization](features.md#synchronization) so that EarlGrey automatically waits
+for elements to appear. As a work around for when that's not possible, `GREYCondition` and `waitWithTimeout`
+are available. The following is an example of waiting for a collection view to populate.
+
+```swift
+let populated = GREYCondition(name: "Wait for UICollectionView to populate", block: { _ in
+    var errorOrNil: NSError?
+
+    EarlGrey().selectElementWithMatcher(collectionViewMatcher)
+              .assertWithMatcher(grey_notNil(), error: &errorOrNil)
+
+    let success = errorOrNil == nil
+    return success
+}).waitWithTimeout(20.0)
+
+GREYAssertTrue(populated, reason: "Failed to populate UICollectionView in 20 seconds")
+```


tirodkar:
LGTM from my side.

docs/faq.md
```diff
@@ -241,3 +241,24 @@ grey_allOf(grey_sufficientlyVisible(),
 then all elements in the entire application will be checked for visibility, and finally one
 with a matching id will be selected. It's significantly faster to use the most targeted
 matchers first (typically `grey_accessibilityID` or `grey_accessibilityLabel`).
+
+**How do I wait for an element to appear using Swift?**
```
khandpur: This one is interesting because you wouldn't need to add explicit waits/sleeps if synchronization is setup correctly. We should mention that this is a workaround and not a best practice.

bootstraponline: Updated.

bootstraponline:
Regarding the error pointer, I think we can just do this:

```swift
GREYCondition(name: "Wait for collection view to populate", block: { _ in
    var errorOrNil : NSError?

    EarlGrey().selectElementWithMatcher(collectionViewMatcher)
              .assertWithMatcher(grey_notNil(), error: &errorOrNil)

    let success = errorOrNil == nil
    return success
}).waitWithTimeout(20.0)
```

That's much cleaner than dealing with `UnsafeMutablePointer`

docs/faq.md
```diff
@@ -266,3 +266,21 @@ Alternatively, perform a runtime check for the EarlGrey class:
 // Swift
 public static let envEarlGrey:Bool = NSClassFromString("EarlGreyImpl") != nil
 ```
+
+**How do I wait for an element to appear using Swift?**
+
+The best way is to [setup synchronization](features.md#synchronization) so that EarlGrey automatically waits
+for elements to appear. As a work around for when that's not possible, `GreyCondition` and `waitWithTimeout`
```
khandpur: `GreyCondition` => `GREYCondition`

docs/faq.md
```diff
@@ -266,3 +266,21 @@ Alternatively, perform a runtime check for the EarlGrey class:
...
+```swift
+GREYCondition(name: "Wait for collection view to populate", block: { _ in
+    var errorOrNil: NSError?
+
+    EarlGrey().selectElementWithMatcher(collectionViewMatcher)
+              .assertWithMatcher(grey_notNil(), error: &errorOrNil)
+
+    let success = errorOrNil == nil
+    return success
+}).waitWithTimeout(20.0)
```
khandpur: waitWithTimeout returns a success or failure in form of YES/NO. That return value should checked.

bootstraponline: checked and then what?

khandpur: a failure (NO) should be either logged or you can assert:
`GREYAssertTrue(retVal, @"Failed to populate Collection View in 20 seconds")`

bootstraponline: Makes sense. Updated.

docs/faq.md

khandpur:
LGTM

PR #123, opened 2016-06-07
Add Swift action example

### LINKED ISSUE #111: Custom action Swift example
It'd be nice to have a simple example of how to do a custom action using Swift.

bootstraponline:
Fix #111

### DIFF
--- a/Tests/FunctionalTests/Sources/EarlGrey.swift
+++ b/Tests/FunctionalTests/Sources/EarlGrey.swift
@@ -14,6 +14,24 @@
 // limitations under the License.
 //
 
+func grey_replaceText(text: String) -> GREYAction {
+  return GREYActionBlock.actionWithName("Replace text \"\(text)\"", constraints: grey_not(grey_systemAlertViewShown()),
+                                        performBlock: { element, errorOrNil in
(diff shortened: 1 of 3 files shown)


tirodkar:
I've got a different action that I'm adding to the basic views test here. Also, could you create a local GREYActionBlock and use it instead of returning it directly?

bootstraponline:
> I've got a different action that I'm adding to the basic views test here.

Is that suppose to be linked?

bootstraponline:
> Also, could you create a local GREYActionBlock and use it instead of returning it directly?

So remove the wrapper function and just keep `GREYActionBlock.actionWithName`? Or keep the wrapper & add an example usage?

tirodkar:
Remove the wrapper function. Just have the block.

I'd think it would be nicer to see an example that is also in a test, so it can be immediately referenced.

bootstraponline:
> I'd think it would be nicer to see an example that is also in a test, so it can be immediately referenced.

I agree. I added this one to `FTRSwiftTests.swift`. Let me know what you think.

bootstraponline:
closing in favor of tirodkar's upcoming PR.

PR #130, opened 2016-06-09
Update AX=N matcher docs

bootstraponline:


docs/faq.md
```diff
@@ -311,4 +311,9 @@ let populated = GREYCondition(name: "Wait for UICollectionView to populate", blo
 }).waitWithTimeout(20.0)
 
 GREYAssertTrue(populated, reason: "Failed to populate UICollectionView in 20 seconds")
-```
\ No newline at end of file
+```
+
+**How do I match AX=N elements?**
+
+Elements with accessibility disabled (`AX=N`) can still be matched by `grey_accessibilityID`. The other accessibility matchers (`grey_accessibilityLabel`, `grey_accessibilityValue`, `grey_accessibilityTrait`, `grey_accessibilityHint`) won't work. If the `AX=N` element can't be matched by id, then
```
khandpur: AX=N is a specific thing in the UI hierarchy. Maybe add that it's present in the UI hierarchy only.

khandpur: "If the `AX=N` element can't be matched by id" => perhaps you should also suggest that accessibility ID is something that can be added to non-accessible elements.

bootstraponline:
Updated.

docs/faq.md
```diff
@@ -311,4 +311,18 @@ let populated = GREYCondition(name: "Wait for UICollectionView to populate", blo
 }).waitWithTimeout(20.0)
 
 GREYAssertTrue(populated, reason: "Failed to populate UICollectionView in 20 seconds")
-```
\ No newline at end of file
+```
+
+**How do I match AX=N elements?**
```
khandpur: **How do I match AX=N elements?** =>
**How do I match elements that are denoted with "AX=N" in the view hierarchy?**

Issue #132, opened 2016-06-09
Add examples.md doc


khandpur:
FAQ has tons of examples that should be moved to this doc.

khandpur:
Moved most questions from FAQ to stack overflow. Examples.md should be added next.

PR #135, opened 2016-06-10
Add grey_firstElement() to examples

bootstraponline:
#132

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -1,0 +1,39 @@
+**Selecting the first match via a custom Swift matcher**
+
+When an application has exact duplicate elements (every attribute is the same,
+including their hierarchy), then the correct fix is to update the app to avoid
+duplicating the UI. When that's not possible, a workaround is to match on
+the first element.
+
+```swift
+// Swift Custom Matcher
+/*
+ *  Example Usage:
+ *
+ *  EarlGrey().selectElementWithMatcher(grey_allOfMatchers(
+ *    grey_accessibilityID("some_id"),
+ *    grey_interactable(),
+ *    grey_firstElement()))
+ *  .assertWithMatcher(grey_notNil())
+ *
+ *  Only intended to be used with selectElementWithMatcher. Don't
+ *  use grey_firstElement in assertWithMatcher as it won't do anything.
+ */
+func grey_firstElement() -> GREYMatcher {
+  var firstMatch = true
+  let matches: MatchesBlock = { (element: AnyObject!) -> Bool in
+    if firstMatch {
+      firstMatch = false
+      return true
+    }
+
+    return false
+  }
+
+  let description: DescribeToBlock = { (description: GREYDescription!) -> Void in
+    description.appendText("first match")
+  }
+
+  return GREYElementMatcherBlock.init(matchesBlock: matches, descriptionBlock: description)
+}
+```


docs/examples.md
```diff
@@ -0,0 +1,41 @@
...
+    if firstMatch {
+      firstMatch = false
+      return true
+    }
+
+    return false
+  }
+
+  let description: DescribeToBlock = { (description: GREYDescription!) -> Void in
+    guard let description = description else {
```
bootstraponline: Why is description sometimes nil?

khandpur: it can be empty but shouldn't be nil. You're supposed to append matcher's description to it.

bootstraponline: I'm getting exceptions when running `description.appendText("first match")` without the guard.

khandpur: ah, i see yes it can be nil...
https://github.com/google/EarlGrey/blob/0481d29a250a26f7281bad2b5b2019a3060a8c82/EarlGrey/Matcher/GREYAnyOf.m#L37

We should fix that.

bootstraponline: Nice, I agree! Opened https://github.com/google/EarlGrey/issues/136

tirodkar: I think we should close this or find another way to provide it or so until we fix that issue.

bootstraponline: I think it'd be better to merge since this is what people have to do right now? Then update once the bug is fixed.

sid-github: Update this as https://github.com/google/EarlGrey/pull/137 is in?

tirodkar:
LGTM

docs/examples.md
tirodkar: @bootstraponline, should the guard be removed now that your fix is in?

bootstraponline: > @bootstraponline, should the guard be removed now that your fix is in?

yes, updated.

docs/examples.md
```diff
@@ -0,0 +1,39 @@
+**Selecting the first match via a custom Swift matcher**
+
+When an application has exact duplicate elements (every attribute is the same,
+including their hierarchy), then the correct fix is to update the app to avoid
+duplicating the UI. When that's not possible, a workaround is to match on
+the first element.
```
khandpur: Instead of the first element, couldn't we generalize this to match on Nth element?
I believe that can be baked in to the API :)

docs/examples.md
```diff
@@ -0,0 +1,39 @@
...
+
+```swift
+// Swift Custom Matcher
+/*
+ *  Example Usage:
+ *
+ *  EarlGrey().selectElementWithMatcher(grey_allOfMatchers(
+ *    grey_accessibilityID("some_id"),
+ *    grey_interactable(),
+ *    grey_firstElement()))
```
khandpur: worth mentioning that order of matcher is important (most -> least selective) and grey_firstElement should come last since it is the least selective matcher.

khandpur:
@bootstraponline - I feel that this would serve better as a first party API. Thoughts?

docs/examples.md
bootstraponline: yeah, that's a great idea. I received similar feedback internally.

bootstraponline:
> @bootstraponline - I feel that this would serve better as a first party API. Thoughts?

agreed

bootstraponline:
Closing in favor of implementing Nth element matcher in EarlGrey core.

https://github.com/google/EarlGrey/issues/138

PR #137, opened 2016-06-13
Fix nil condition Swift error

### LINKED ISSUE #136: Fix description is sometimes nil
@khandpur found the code responsible:
https://github.com/google/EarlGrey/blob/0481d29a250a26f7281bad2b5b2019a3060a8c82/EarlGrey/Matcher/GREYAnyOf.m#L37

bootstraponline:
Fix #136

### DIFF
--- a/EarlGrey/Matcher/GREYAllOf.m
+++ b/EarlGrey/Matcher/GREYAllOf.m
@@ -16,6 +16,7 @@
 
 #import "Matcher/GREYAllOf.h"
 #import "Matcher/GREYDescription.h"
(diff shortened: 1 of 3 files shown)


tirodkar:
LGTM from my side.

sid-github:
Hi, thanks for the fix! Can you please provide more information on the Swift error you are fixing here? Like a log output? I feel that this may be fixed by adding nullability information to https://github.com/google/EarlGrey/blob/master/EarlGrey/Matcher/GREYMatcher.h#L47 instead of passing in a description object that is not used.

Issue #138, opened 2016-06-15
Add grey_atIndex(...) element matcher


bootstraponline:
Useful for correctly resolving duplicate element issues by index (picking first, last, etc.).

Background: https://github.com/google/EarlGrey/pull/135

`grey_atIndex(...)`

Use cases

- Find nth element in a table
- Resolve duplicate match

TyrantDante:
What is the sequence

PR #119, opened 2016-06-07
Add grey_replaceText

bootstraponline:
Fix #110

**To Do**

- [x] Reuse `actionForClearText` webview logic
- [x] Add webview replaceText test

PR #139, opened 2016-06-15
Add EarlGreySwiftUnitTests target to UnitTests project

bootstraponline:


bootstraponline:
discussed on slack: decided to move the example code to a new testing target in the demo app so we can verify the code continues to work as the APIs evolve.

To Do

- [ ] move example code to demo app
- [ ] update examples.md to reference code in demo app

bootstraponline:
I updated the example project with a `EarlGreyContribSwiftTests` test target that contains the custom Swift matchers. `examples.md` now links to this file directly.

Demo/EarlGreyExample/EarlGreyContribSwiftTests/GetText.swift
```diff
@@ -0,0 +1,22 @@
...
+ *  Example Usage:
+ *
+ *  let element = Element()
+ *
+ *  domainField.performAction(grey_typeText("hello.there"))
+ *             .performAction(grey_getText(element))
+ *
+ *  GREYAssertTrue(element.text != "", reason: "get text failed")
+ */
+func grey_getText(elementCopy: Element) -> GREYActionBlock {
```
tirodkar: Is there any reason why these are Swift only classes? We can add these to GREYActions or so and have them exported in the exact same way that all other matchers are. 

Also, I'm not completely sure we need an entirely new project for this. In the FunctionalTests.xcodeproj, could we not add the view controller and tests for these methods with a separate of "EarlGreyExperimentalTests.m" file or so?

bootstraponline: > Is there any reason why these are Swift only classes?

Yes, I have a growing collection of Swift examples. There are already a bunch of examples for Objective C. Before the code would just be in the markdown file, instead now they're in a test target that properly validates the functionality of the examples.

> We can add these to GREYActions or so and have them exported in the exact same way that all other matchers are.

That's the long term plan ([for example: grey_replaceText](https://github.com/google/EarlGrey/pull/119)). I expect grey_firstElement will be replaced by an [nth element matcher](https://github.com/google/EarlGrey/issues/138) in core once that's implemented.

`grey_getText` is new and it seemed safer to let people try it out first before adding a similar matcher to EarlGrey core. Ideally all the matchers would end up in `GREYActions` once they've proven useful.

> Also, I'm not completely sure we need an entirely new project for this.

To clarify, this is a new testing target on an existing project -- not a new project. 

> In the FunctionalTests.xcodeproj, could we not add the view controller and tests for these methods with a separate of "EarlGreyExperimentalTests.m" file or so?

`FunctionalTests` test EarlGrey itself and the new matchers aren't intended to be accepted into EarlGrey yet. If it's fine to just add the experimental matchers directly to core and test them like all the other features, then `FunctionalTests` makes sense. Do we have a way to mark APIs as experimental? I expect them to change (potentially quite a bit).

tirodkar: > FunctionalTests test EarlGrey itself and the new matchers aren't intended to be accepted into EarlGrey yet. If it's fine to just add the experimental matchers directly to core and test them like all the other features, then FunctionalTests makes sense. Do we have a way to mark APIs as experimental? I expect them to change (potentially quite a bit).

FunctionalTests just tests any implemented matchers. The demo project is simply there for setup questions and shouldn't really be looked at for matcher etc. questions. I think we should add it to the FunctionalTests project since we already have Matcher questions / goto examples there. Let's keep this waiting then till `grey_firstElement` is supplanted by `grey_nthElement`.

Thanks a lot for the contributions!

bootstraponline:
discussed on slack, will be revising the PR using a new approach:

- `EarlGreyExamples` will be for examples of existing APIs 
- `FunctionalTest` will have a new target for experimental new APIs

tirodkar:
Thanks @bootstraponline, keepign this open till you update it.

bootstraponline:
Updated. I think the new approach makes a lot more sense. Thanks for the feedback @tirodkar.

bootstraponline:
Also FunctionalTest is a mess (super long & flaky) so I ended up using the UnitTest project instead.

tirodkar:
There's some expected merge conflicts here. Secondly, we should be showing example with UI. What seems to be the messy part here? Are the flakes ones we know about already?

bootstraponline:
> Secondly, we should be showing example with UI.

I think we should show examples only for existing APIs, not experimental ones.

> What seems to be the messy part here? Are the flakes ones we know about already?

The tests don't pass and they take a long time to run. I'd much rather rely on Unit Tests to identify breaks in the experimental APIs. For examples of already existing APIs, I agree functional testing makes more sense.

bootstraponline:
I fixed the merge conflict.

bootstraponline:
Discussed on Slack. Will be moving to a contribs xcode project instead of a new testing target.

PR #148, opened 2016-06-19
Simplify code and make it compatible with iOS 10.0 beta 1

axi0mX:
Simplify code and make it compatible with iOS 10.0 beta 1.
1) enabling accessibility
2) synchronization with UIGestureRecognizers

Tested on:
8.1 simulator on Xcode 7.3.1
8.1.3 device on Xcode 7.3.1
9.3 simulator on Xcode 7.3.1
9.3.1 device on Xcode 7.3.1
10.0 beta 1 simulator on Xcode 8 beta 1
10.0 beta 1 device on Xcode 8 beta 1

Reboot is required after installing Xcode 8 beta 1 before AX works in simulator. This is not an issue with my code, it also affects Apple's new AX Inspector.


tirodkar:
@axi0mX 

> "Gesture recognition code needs a separate fix for iOS 10.0 beta 1; to test this pull request just rename method +[_UIGestureRecognizerFailureMap_GREYAdditions load] to prevent it from executing"

Can't we also provide a fix for that somehow? Seems to me like this still doesn't fix the entire issue. Thanks a lot btw.

axi0mX:
It's a separate issue, let's keep it out of scope.

EarlGrey/Core/GREYAutomationSetup.m
```diff
@@ -45,37 +45,29 @@ - (void)perform {
...
-  NSString *accessibilitySettingsPrefBundlePath =
-      @"/System/Library/PreferenceBundles/AccessibilitySettings.bundle/AccessibilitySettings";
-  NSString *accessibilityControllerClassName = @"AccessibilitySettingsController";
-  id accessibilityControllerInstance =
-      [self grey_settingsClassInstanceFromBundleAtPath:accessibilitySettingsPrefBundlePath
-                                         withClassName:accessibilityControllerClassName];
-  [accessibilityControllerInstance setAXInspectorEnabled:@(enabled) specifier:nil];
+// Enables accessibility, which is required for using any properties of the accessibility tree.
++ (void)grey_setEnableAccessibility {
+  CFPreferencesSetAppValue((CFStringRef)@"ApplicationAccessibilityEnabled",
```
tirodkar: Is there a similar solution for Autocorrection as well?

axi0mX: There doesn't seem to be. The code that handles keyboard settings is fairly complicated, and while it may also use CFPreferences internally to set some values, it does a lot of other work as well. Using a higher-level private API that performs that work, the way EarlGrey does it today, seems best. If there are no issues with the code, I'd recommend not changing it.

tirodkar:
@axi0mX, I tested this out but this didn't seem to work for me. Were you able to get the tests running on your simulator?

khandpur:
Didn't work for me either on iOS 10.0 iPhone SE simulator. Tests just freeze.

axi0mX:
It seems that rebooting after installing Xcode 8 beta 1 fixes that issue. No joke.

PR updated with changes for gesture recognizers.

EarlGrey/Synchronization/GREYAppStateTracker.m
```diff
@@ -181,8 +181,8 @@ - (NSString *)grey_stringFromState:(GREYAppState)state {
     [eventStateString addObject:stateMsg];
   }
   if (state & kGREYPendingGestureRecognition) {
-    [eventStateString addObject:@"Waiting for system-wide gesture recognizer to detect or fail a "
```
axi0mX: Removed because kGREYPendingGestureRecognition isn't only for system-wide gesture recognizers, it is for all of them.

axi0mX:
The fix I pushed 6 hours ago had some issues with gesture recognizers involving the keyboard; I just updated this PR to fix that issue.

mbaxley:
Regarding having to reboot for Xcode 8, we've seen this also. NSUserDefaults has some problems in Xcode 8 beta 1, requiring a Mac reboot. Anecdotally, we haven't seen any problems after doing a reboot.
https://forums.developer.apple.com/message/143878#143878

tirodkar:
The reboot did fix the issue. @axi0mX, could you please fix any merge conflicts?

axi0mX:
Merge conflicts fixed.

EarlGrey/Core/GREYAutomationSetup.m
```diff
@@ -45,37 +45,25 @@ - (void)perform {
...
-  NSString *accessibilityControllerClassName = @"AccessibilitySettingsController";
-  id accessibilityControllerInstance =
-      [self grey_settingsClassInstanceFromBundleAtPath:accessibilitySettingsPrefBundlePath
-                                         withClassName:accessibilityControllerClassName];
-  [accessibilityControllerInstance setAXInspectorEnabled:@(enabled) specifier:nil];
+// Enables accessibility, which is required for using any properties of the accessibility tree.
++ (void)grey_setEnableAccessibility {
+  char const *const libAccessibilityPath =
+      [@"/usr/lib/libAccessibility.dylib" fileSystemRepresentation];
+  void *handle = dlopen(libAccessibilityPath, RTLD_LOCAL);
```
axi0mX: I made a minor change here to dlopen/dlsym calls. 

With RTLD_LOCAL, if the dylib wasn't already loaded in global scope, the dynamic linker will not make its symbols available to anyone else. These symbols will then only be accessible through the handle provided by dlopen.

This way is better, because users won't be relying on EarlGrey dlopening this library, and it will be safe to remove this code in the future.

PR #185, opened 2016-07-12
Run loop spinner wake-up timer should first fire after maxSleepInterval

axi0mX:
1) Run loop spinner wake-up timer should first fire after maxSleepInterval
2) Prevent the need for CFBridgingRetain+CFRelease and use ARC

### DIFF
--- a/EarlGrey/Synchronization/GREYRunLoopSpinner.m
+++ b/EarlGrey/Synchronization/GREYRunLoopSpinner.m
@@ -293,12 +293,13 @@ static void (^noopTimerHandler)(CFRunLoopTimerRef timer) = ^(CFRunLoopTimerRef t
  */
 - (CFRunLoopTimerRef)grey_setupWakeUpTimerInMode:(NSString *)mode {
   if (self.maxSleepInterval > 0) {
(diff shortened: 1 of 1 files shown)


EarlGrey/Synchronization/GREYRunLoopSpinner.m
```diff
@@ -294,15 +291,16 @@ - (CFRunLoopObserverRef)grey_setupObserverInMode:(CFStringRef)mode
  *
  *  @return The registered timer or @c nil if no timer was added to @c mode.
  */
-- (CFRunLoopTimerRef)grey_setupWakeUpTimerInMode:(CFStringRef)mode {
+- (CFRunLoopTimerRef)grey_setupWakeUpTimerInMode:(NSString *)mode {
   if (self.maxSleepInterval > 0) {
```
khandpur: extract self.maxSleepInterval into a var because it's used many times

axi0mX: It's an ivar, so I went ahead and changed it to be accessed directly and made the same change for other ivars in this file.

khandpur:
LGTM

PR #186, opened 2016-07-13
Fix usage of EarlGrey.swift file in the EarlGrey repo.

tirodkar:
Issue: Multiple instances of EarlGrey.swift used by multiple projects in the EarlGrey report.
Solution: Make all projects point to EarlGrey.swift in the gem.


tirodkar:
Updated this PR - reverted the removal of EarlGrey.swift from the carthage_after and cocoapods_after projects since after running the tests, the gem copies the EarlGrey.swift file there and compares it against what's present there. These files should not be removed.

bootstraponline:
> Updated this PR - reverted the removal of EarlGrey.swift from the carthage_after and cocoapods_after projects since after running the tests, the gem copies the EarlGrey.swift file there and compares it against what's present there. These files should not be removed.

Are you sure? I just deleted them and the tests still pass (`rake spec`).

bye_earlgrey swift

I think we should just delete the contents but keep the file reference. This way the test can assert the file exists but doesn't have to be updated every time the Swift code changes.

tirodkar:
> Are you sure? I just deleted them and the tests still pass (rake spec).

As you mentioned, I deleted the files and changes the value in the xcodeproj as well. Thanks for the heads up on the deletion. I'll delete the files but keep the projects intact.

khandpur:
LGTM

Issue #162, opened 2016-06-27
If parent directory has spaces, `setup-earlgrey.sh` will fail and exit


danshao:
For example:

EarlGrey is cloned into `/Users/b/Happy Bear/`

```
$ sudo sh Scripts/setup-earlgrey.sh 
Changing into EarlGrey Directory
Scripts/setup-earlgrey.sh: line 199: pushd: /Users/b/Happy: No such file or directory
Obtaining the fishhook dependency
Obtaining the fishhook dependency.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   119    0   119    0     0     68      0 --:--:--  0:00:01 --:--:--    68
100  7810  100  7810    0     0   2683      0  0:00:02  0:00:02 --:--:-- 11046
replace fishhook-0.2/LICENSE? [y]es, [n]o, [A]ll, [N]one, [r]ename: ^CThere was an issue while unzipping the Fishhook zip file. Please ensure if it unzips manually since it might be corrupt.
```

PR #187, opened 2016-07-13
Fix setup script for handling spaces in the directory path.

tirodkar:
Issue: The setup-earlgrey-script throws errors when the directory path has spaces.
Solution: Quote all the paths in setup-earlgrey-script. 
                Fix https://github.com/google/EarlGrey/issues/162


khandpur:
LGTM

PR #192, opened 2016-07-14
Refactor how failure description is generated in default failure handler and XCTestCase additions

axi0mX:
Refactoring, with minor changes in log output.

Improves how failure description is generated in default failure handler and XCTestCase additions.

### DIFF
--- a/EarlGrey/Exception/GREYDefaultFailureHandler.m
+++ b/EarlGrey/Exception/GREYDefaultFailureHandler.m
@@ -52,11 +52,12 @@ static NSUInteger gUnknownTestExceptionCounter = 0;
   }
 
   NSMutableString *log = [[NSMutableString alloc] init];
-  // Start on fresh new line.
-  [log appendString:@"\n"];
-  [log appendFormat:@"Exception: %@\n", exception.name];
   NSString *reason =
       exception.reason.length > 0 ? exception.reason : @"exception.reason was not provided";
+  [log appendFormat:@"%@\n\n", reason];
+  [log appendFormat:@"Bundle ID: %@\n\n", [[NSBundle mainBundle] bundleIdentifier]];
+  [log appendFormat:@"Stack trace: %@\n\n", [NSThread callStackSymbols]];
+  [log appendFormat:@"Exception: %@\n", exception.name];
   [log appendFormat:@"Reason: %@\n", reason];
   if (details.length > 0) {
     [log appendFormat:@"%@\n", details];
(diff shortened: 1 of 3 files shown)


tirodkar:
@axi0mX, while there doesn't seem to be a direct conflict here, this needs to be rebased.

axi0mX:
Done

EarlGrey/Exception/GREYDefaultFailureHandler.m
```diff
@@ -44,18 +44,49 @@ - (void)setInvocationFile:(NSString *)fileName andInvocationLine:(NSUInteger)lin
 
 - (void)handleException:(GREYFrameworkException *)exception details:(NSString *)details {
```
khandpur: Can you please explain in detail why are you making this change? What's this going to achieve? If all you need is some additional fields printed, just include those as changes.

EarlGrey/Exception/GREYDefaultFailureHandler.m
```diff
@@ -52,11 +52,12 @@ - (void)handleException:(GREYFrameworkException *)exception details:(NSString *)
   }
 
   NSMutableString *log = [[NSMutableString alloc] init];
-  // Start on fresh new line.
-  [log appendString:@"\n"];
-  [log appendFormat:@"Exception: %@\n", exception.name];
   NSString *reason =
       exception.reason.length > 0 ? exception.reason : @"exception.reason was not provided";
+  [log appendFormat:@"%@\n\n", reason];
+  [log appendFormat:@"Bundle ID: %@\n\n", [[NSBundle mainBundle] bundleIdentifier]];
```
sid-github: I am not sure how bundle ID would be useful in the logs given that is constant throughout test execution and not used for debugging any thing. Please remove if you think so too.

tirodkar: It would help in case of looking at the app / test in case of multiprocess if I'm right.

axi0mX: Let's keep it.

In cases where someone copy and pastes the test log or a part of the test log it may not be clear which app bundle the snippet is from, especially when the development and production copy of the app have a different bundle ID so you can have both on your phone at the same time.

Furthermore, in the multiprocess model log output needs to have a bundle ID so you know which app generated the error message.

sid-github: SGTM, yup lets keep it, I did not think about multi-process.

EarlGrey/Exception/GREYDefaultFailureHandler.m
```diff
@@ -52,11 +52,12 @@ - (void)handleException:(GREYFrameworkException *)exception details:(NSString *)
   }
 
   NSMutableString *log = [[NSMutableString alloc] init];
-  // Start on fresh new line.
-  [log appendString:@"\n"];
```
sid-github: Not starting on new line anymore? Wouldn't new line be cleaner?

axi0mX: No. Log output which was put into a format string in XCTestCase+GREYAdditions is now in this file.

A new line was used here before because it put a blank line between different strings being put in a format string in XCTestCase+GREYAdditions, but not in front of the first string.

The first string originally in XCTestCase+GREYAdditions never had a blank line in front of it. It must not have one, because the first line is shown as explanation of error by Xcode, and if it only has a new line the message will be blank. (Example of this message: http://iosunittesting.com/wp-content/uploads/TestAssistantExample.png)

PR #196, opened 2016-07-21
Fixing some minor styling issue and how assertions are propagated as failures. Also removing NS_BLOCK_ASSERTION.

khandpur:
1) Fixing some minor styling issue
2) Failure outside valid test context now throw exception instead of silently ignoring the failures
3) Removing NS_BLOCK_ASSERTION so EarlGrey can be used again release builds.


tirodkar:
LGTM from me!

sid-github:
LGTM too.

Issue #50, opened 2016-03-21
Travis is failing


bootstraponline:
I updated [the docs](https://github.com/google/EarlGrey/pull/49) and was surprised to see [all the Travis failures.](https://travis-ci.org/google/EarlGrey/builds/116943764)
## 

> The job exceeded the maxmimum time limit for jobs, and has been terminated.
> 
> https://travis-ci.org/google/EarlGrey/jobs/116943768
## 

> Executed 208 tests, with 2 failures (2 unexpected) in 2361.286 (2361.643) seconds
> 
> https://travis-ci.org/google/EarlGrey/jobs/116943769

tirodkar:
Thanks for bringing this to our attention @bootstraponline, we're currently working on fixing these issues.

bootstraponline:
Travis has a [hard limit of 50 minutes for Mac builds.](https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts) If there was some way to cache the app build instead of recompiling every time, that might help.

For appium, we've had to use an open source cloudbees account with [on premise executors](https://documentation.cloudbees.com/docs/dev-at-cloud/On-Premise+Executors.html) (via AWS) because Travis doesn't play nice with lots of UI tests.

khandpur:
Yeah it's been pretty annoying trying to work around that. Also, Travis has a log size limit of 2 (or 3) mb I don't quite recall exactly. I switched to using xcpretty for that reason however I think xcpretty sometimes hangs or at least doesn't output stuff which makes travis believe xcodebuild is stuck for more than 10 mins and it terminates our test run.

bootstraponline:
Yeah, another awesome feature of Travis:

> If no output is received from a build for 10 minutes, it’s assumed to have stalled for unknown reasons and is subsequently killed.
> https://docs.travis-ci.com/user/common-build-problems/

Probably why [Swift provides their own hardware for CI](https://swift.org/blog/swift-ci/). Hopefully Google has some way of provisioning Macs for builds. I wish there was an AWS like service for Mac builds (without silly timeouts).

bootstraponline:
[Travis uses MacStadium](http://www.macstadium.com/blog/how-travis-ci-halved-boot-times-for-os-x-and-ios-testing-at-macstadium/). Using them directly would mean no limits on the builds. For CI, it'd be easy to plug into a free cloudbees jenkins account.

tirodkar:
Does this still need to be open?

[...]

PR #202, opened 2016-07-28
Introducing retries if xcodebuild fails before tests are even run.

khandpur:
If xcodebuild exits with code 65 and tests haven't been started it, retry up to 3 times in total. There is a possibility that due to compilation issue, error code 65 is returned, in which case all retries will fail with the same error code but it should be fairly inexpensive to do that.

Also redirecting all output to xcodebuild.log which will be printed on failure. We may want to upload the logs to some dashboard after failure but that's scoped for future.

### DIFF
--- a/Scripts/travis.sh
+++ b/Scripts/travis.sh
@@ -35,8 +35,13 @@ execute_xcodebuild() {
   fi
 
   for retry_attempts in {1..3}; do
+    # To retry on failure, disable exiting if command below fails.
+    set +e
     env NSUnbufferedIO=YES xcodebuild -project ${1} -scheme ${2} -sdk "$SDK" -destination "$DESTINATION" -configuration Debug ONLY_ACTIVE_ARCH=NO test | tee xcodebuild.log | xcpretty -s;
-    if [ $? -ne 65 ]; then
+    retval=$?
+    # Re-enable exiting for command failures.
+    set -e
+  if [ ${retval} -ne 65 ]; then
       break
     fi
   done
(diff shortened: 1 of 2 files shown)


Scripts/travis.sh
```diff
@@ -19,22 +19,59 @@ set -euxo pipefail
...
+    exit 1
+  fi
+
+  retval_xcodebuild=0
+  for retry_attempts in {1..3}; do
+    # To retry on failure, disable exiting if command below fails.
+    set +e
+    env NSUnbufferedIO=YES xcodebuild -project ${1} -scheme ${2} -sdk "$SDK" -destination "$DESTINATION" -configuration Debug ONLY_ACTIVE_ARCH=NO test | tee xcodebuild.log | xcpretty -s;
+    retval_xcodebuild=$?
+    # Even failed tests exit with code 65. Add a check to ensure test haven't started.
```
tirodkar: nit: tests haven't started

sid-github: nit: Either Aditya's comment or:
# Even failed tests exit with code 65, check to ensure tests haven't started.

Scripts/travis.sh
```diff
@@ -19,22 +19,59 @@ set -euxo pipefail
...
+    # we achieve that by looking for keyword "Test Suite" in xcodebuild.log.
```
sid-github: *We
also, add a TODO for a better check as this check is brittle and can fail when logging changes.

khandpur:
Comments addressed

khandpur:
Part of https://github.com/google/EarlGrey/issues/50

PR #206, opened 2016-08-03
Internal to external sync - 08/03/2016

tirodkar:
What this does:

* Adds an analytics Proxy change to not rely on the order of tests.
* Enables chaining of exception and signal handlers. You can now reraise the same signal with default handlers in places where before, you'd see a stack dump.


khandpur:
LGTM

Issue #212, opened 2016-08-09
Add support for EarlGrey.swift and Swift 2.2


tirodkar:
We updated our EarlGrey.swift file to 3 but this acts as a blocker for anyone not using Swift 3 that could be easily overcome by having two version of the EarlGrey.swift file for Swift 2.2 and Swift 3.

PR #213, opened 2016-08-09
Add support for a 2.2 EarlGrey.swift version in the EarlGrey gem.

tirodkar:
This still needs to be pushed out to the official gem. It worked perfectly fine locally.

Fixes https://github.com/google/EarlGrey/issues/212


gem/lib/earlgrey/cli.rb
```diff
@@ -19,11 +19,12 @@ class CLI < Thor
     package_name 'EarlGrey'
 
     desc 'install', 'Installs EarlGrey into an Xcode unit test target'
+    method_option :carthage, type: :boolean, default: true
     method_option :project, aliases: '-p', type: :string, required: false, desc: 'Project'
-    method_option :target,  aliases: '-t', type: :string, required: true, desc: 'EarlGreyTestTarget'
     method_option :scheme,  aliases: '-s', type: :string, required: false, desc: 'EarlGreyTestTarget.xcscheme'
     method_option :swift,    type: :boolean, default: true
-    method_option :carthage, type: :boolean, default: true
+    method_option :swift_version, type: :numeric, default: 3.0, :enum => %w(2.2, 3.0)
```
bootstraponline: Numerics in Thor are parsed as floats or integers. This doesn't work for point releases of Swift. For example, the upcoming [Swift `3.0.1`](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160725/025676.html).

The correct datatype is string.

bootstraponline:
```ruby
%w(2.2, 3.0)
```

The above creates a corrupted first version value. Observe:

```
> %w(2.2, 3.0)
=> ["2.2,", "3.0"]
```

There shouldn't be a comma in element 0. The space delimiter is what's used, not a comma.

gem/lib/earlgrey/cli.rb
```diff
@@ -36,7 +37,7 @@ def install
 
       # CLI will never use Cocoapod's `post_install do |installer|`
       podfile_installer = nil
-      opts = { swift: o[SWIFT], carthage: o[CARTHAGE] }
+      opts = { swift: o[SWIFT], carthage: o[CARTHAGE], swift_version: o[SWIFT_VERSION] }
```
bootstraponline: I don't see `SWIFT_VERSION` defined anywhere in this pull request?

bootstraponline: Also this is why the Ruby tests are failing:

>  uninitialized constant EarlGrey::CLI::SWIFT_VERSION
>
> https://travis-ci.org/google/EarlGrey/jobs/150811297

gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -66,6 +66,7 @@ def puts_yellow(string)
     def set_defaults(installer, project_name, test_target_name, scheme_file, opts = {})
       @swift = opts.fetch(:swift, false)
       @carthage = opts.fetch(:carthage, false)
+      @swift_version = opts.fetch(:swift_version, false)
```
bootstraponline: False is not a valid default value for `:swift_version`. `:swift` and `:carthage` are booleans which is why they default to false. `:swift_version` should default to a valid version of Swift.

gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -353,7 +354,8 @@ def copy_swift_files
       dst_header = File.join(dst_root, src_header_name)
 
       src_swift_name = 'EarlGrey.swift'
-      src_swift = File.join(src_root, src_swift_name)
+      src_swift = File.join(src_root, "Swift-" + "#{swift_version}/" + src_swift_name)
```
bootstraponline: When using interpolation (`#{}`), you don't use +. Also the path separator should never be hard coded as that's the reason we use the Join method. See the corrected code below.

```ruby
 File.join(src_root, "Swift-#{swift_version}", src_swift_name)
```

bootstraponline:
There are no specs. I think at a minimum, we want specs that verify the correct version of EarlGrey.swift is copied over for:

- The default case
- Version 2.2
- Version 3.0

gem/lib/earlgrey/cli.rb
tirodkar: Done. Thanks!

gem/lib/earlgrey/cli.rb
```diff
@@ -19,24 +19,26 @@ class CLI < Thor
...
 
     desc 'install', 'Installs EarlGrey into an Xcode unit test target'
+    method_option :carthage, type: :boolean, default: true
     method_option :project, aliases: '-p', type: :string, required: false, desc: 'Project'
-    method_option :target,  aliases: '-t', type: :string, required: true, desc: 'EarlGreyTestTarget'
     method_option :scheme,  aliases: '-s', type: :string, required: false, desc: 'EarlGreyTestTarget.xcscheme'
-    method_option :swift,    type: :boolean, default: true
-    method_option :carthage, type: :boolean, default: true
+    method_option :swift, type: :boolean, default: true
+    method_option :swift_version, type: :string, default: '3.0', enum: ['2.2', '3.0']
```
bootstraponline: This is a cleaner array syntax ([as mentioned in the Thor wiki](https://github.com/erikhuda/thor/wiki/Method-Options):

```ruby
%w(2.2 3.0)
```

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
```diff
@@ -0,0 +1,88 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import EarlGrey
+
+func grey_allOfMatchers(args: AnyObject...) -> GREYMatcher! {
```
bootstraponline: per https://github.com/google/EarlGrey/pull/216 I think these should be public?

gem/spec/configure_earlgrey_spec.rb
```diff
@@ -22,8 +22,22 @@
...
 
-  it 'configures for cocoapods' do
-    diff_project cocoapods_after, %w[install -t AutoEarlGrey --no-carthage]
+  it 'configures for cocoapods with default swift version' do
+    diff_project cocoapods_after_default, %w[install -t AutoEarlGrey --no-carthage]
+    expect(EarlGrey.swift).to eq(true)
+    expect(EarlGrey.carthage).to eq(false)
+  end
+
+  it 'configures for cocoapods with Swift 2.2' do
```
bootstraponline: technically this works however we'll potentially end up with 6 fixture projects

2 (cocoapods + carthage) * 3 (swift default, swift 2, swift 3) = 6 

I'd rather keep one baseline for both cocoapods & carthage then dynamically modify the data in memory with the correct path. The only change is the file path so I think this is reasonable.

bootstraponline:
Actually, I'm not seeing any difference between the fixtures at all.... this passes for me:

```diff
diff --git a/gem/spec/configure_earlgrey_spec.rb b/gem/spec/configure_earlgrey_spec.rb
index e4cdf59..b505959 100644
--- a/gem/spec/configure_earlgrey_spec.rb
+++ b/gem/spec/configure_earlgrey_spec.rb
@@ -29,14 +29,14 @@ describe 'configure_earlgrey' do
   end
 
   it 'configures for cocoapods with Swift 2.2' do
-    diff_project cocoapods_after_swift_2,
+    diff_project cocoapods_after_default,
                  %w[install -t AutoEarlGrey --no-carthage --swift_version=2.2]
     expect(EarlGrey.swift).to eq(true)
     expect(EarlGrey.carthage).to eq(false)
   end
 
   it 'configures for cocoapods with Swift 3' do
-    diff_project cocoapods_after_swift_3,
+    diff_project cocoapods_after_default,
                  %w[install -t AutoEarlGrey --no-carthage --swift_version=3.0]
     expect(EarlGrey.swift).to eq(true)
     expect(EarlGrey.carthage).to eq(false)
```

which makes sense because regardless of the source file name, the destination file name is always `EarlGrey.swift` so there's no change.

tirodkar:
Yeah, the file is still EarlGrey.swift so there shouldn't be any change whatsoever there.

khandpur:
You'd need to rebase this.

gem/lib/earlgrey/cli.rb
tirodkar: That actually threw me an error on `rake`, asking me to move to the hash syntax.

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
tirodkar: Yep! Thanks.

gem/lib/earlgrey/cli.rb
bootstraponline: Oh, that's just a lint check. Use `%w[2.2 3.0]` instead.

In Ruby the token after the w can be anything.

```
%w*a b*
=> ["a", "b"]
```

tirodkar:
Rebased and updated. I removed the extra projects and instead have an extra parameter passed to the project_diff file.

bootstraponline:
> confirm that they're okay with these commits being contributed to Google.

I'm okay with these commits being contributed to Google & I've signed the CLA. 😄

bootstraponline:
Looks good to merge.

khandpur:
resolving CLA status.

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
```diff
@@ -0,0 +1,88 @@
...
+public func GREYAssertTrue(@autoclosure expression: () -> BooleanType, reason: String) {
+  GREYAssert(expression().boolValue,
+    reason,
+    details: "Expected the boolean expression to be true")
+}
+
+public func GREYAssertFalse(@autoclosure expression: () -> BooleanType, reason: String) {
+  GREYAssert(!expression().boolValue,
+    reason,
+    details: "Expected the boolean expression to be true")
```
sid-github: s/true/false

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
```diff
@@ -0,0 +1,88 @@
...
+public func grey_allOfMatchers(args: AnyObject...) -> GREYMatcher! {
+  return GREYAllOf.init(matchers: args)
+}
+
+public func grey_anyOfMatchers(args: AnyObject...) -> GREYMatcher! {
+  return GREYAnyOf.init(matchers: args)
+}
+
+public func EarlGrey() -> EarlGreyImpl! {
+  return EarlGreyImpl.invokedFromFile(__FILE__, lineNumber: __LINE__)
```
sid-github: shouldnt these be ##file  and ##line?
I am comparing with https://github.com/google/EarlGrey/blob/master/gem/lib/earlgrey/files/EarlGrey.swift

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
```diff
@@ -0,0 +1,88 @@
...
+  GREYAssert(expression() != nil, reason, details: "Expected expression to be not nil")
+}
+
+public func GREYAssertNil(@autoclosure expression: () -> Any?, reason: String) {
+  GREYAssert(expression() == nil, reason, details: "Expected expression to be nil")
+}
+
+public func GREYAssertEqual<T : Equatable>(@autoclosure left: () -> T?,
+  @autoclosure _ right: () -> T?, reason: String) {
+    GREYAssert(left() == right(), reason, details: "Expeted left term to be equal to right term")
```
sid-github: *Expected

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
```diff
@@ -0,0 +1,88 @@
...
+
+public func GREYFail(reason: String, details: String) {
+  greyFailureHandler.handleException(GREYFrameworkException(name: kGREYAssertionFailedException,
+    reason: reason),
+    details: details)
+}
+
+private func GREYAssert(@autoclosure expression: () -> BooleanType,
+  _ reason: String, details: String) {
+    //GREYSetCurrentAsFailable()
```
sid-github: why is this commented out?

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
tirodkar: Nope this for 2.2

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
tirodkar: Removing.

tirodkar: Removed.

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
tirodkar: Done.

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
tirodkar: Done.

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
```diff
@@ -0,0 +1,88 @@
...
+}
+
+private func GREYAssert(@autoclosure expression: () -> BooleanType,
+  _ reason: String, details: String) {
+    GREYSetCurrentAsFailable()
+    if !expression().boolValue {
+      greyFailureHandler.handleException(GREYFrameworkException(name: kGREYAssertionFailedException,
+        reason: reason),
+        details: details)
+    }
```
sid-github: indent -2

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
```diff
@@ -0,0 +1,88 @@
...
+  GREYAssert(expression() != nil, reason, details: "Expected expression to be not nil")
+}
+
+public func GREYAssertNil(@autoclosure expression: () -> Any?, reason: String) {
+  GREYAssert(expression() == nil, reason, details: "Expected expression to be nil")
+}
+
+public func GREYAssertEqual<T : Equatable>(@autoclosure left: () -> T?,
+  @autoclosure _ right: () -> T?, reason: String) {
+    GREYAssert(left() == right(), reason, details: "Expected left term to be equal to right term")
```
sid-github: indent -2

tirodkar: Done.

gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
tirodkar: Done.

Issue #214, opened 2016-08-09
cocoapods 0.39 / xcodeproj conflict


gurgeous:
I tried to use EarlGrey with cocoapods 0.39. I think this is supported, but there's a conflict between xcodeproj versions. Pod install fails:

`Gem::LoadError - Unable to activate earlgrey-0.0.8, because xcodeproj-0.28.2 conflicts with xcodeproj (>= 1.0)`

https://rubygems.org/gems/cocoapods/versions/0.39.0
https://rubygems.org/gems/earlgrey/versions/0.0.8

Am I interpreting this correctly? Any suggestions?

bootstraponline:
> I tried to use EarlGrey with cocoapods 0.39. I think this is supported,

It's not. The EarlGrey gem requires xcodeproj 1.0 or newer.

tirodkar:
@gurgeous, CocoaPods has an internal dependency on xcodeproj which the earlgrey gem conflicts with. You can still use EarlGrey with CocoaPods, but you won't be able to use the gem for it.

We **highly** recommend that you update your CocoaPods version. In case you're blocked and cannot do so - take a look at our old EarlGrey install and release guide for CocoaPods where you'd have to manually copy the EarlGrey.swift and configure_earlgrey_pods file.

https://github.com/google/EarlGrey/blob/11a4f329df59e01426235561f9a1d7ff0c10c43e/docs/install-and-run.md

gurgeous:
I will attempt to update to the latest cocoapods. Our app is complex and the upgrade we attempted several months ago was not a success.

We should probably revise the install page:

https://github.com/google/EarlGrey/blob/master/docs/install-and-run.md

which contains stale instructions for using 0.39. Thanks guys!

khandpur:
Thanks for updating our instructions Adam!

PR #215, opened 2016-08-10
remove references to Cocoapods 0.39 (fixes issue #214)

gurgeous:


gurgeous:
I signed it!

khandpur:
Hello, thanks for sending this PR. Could you elaborate on the reason behind it? Is this in reference to https://github.com/google/EarlGrey/issues/214 ?

bootstraponline:
Did you try:

> If you don't want a gem dependency, replace require 'earlgrey' with load 'configure_earlgrey.rb' and manually copy configure_earlgrey.rb from the gem into the same directory as the Podfile.

tirodkar:
@gurgeous, thanks for this, but we don't want to remove the old Podfile section just yet. Remember, you can still use EarlGrey with your 0.38 or so, just without the gem.

I'd recommend adding a change similar to what @bootstraponline stated in Point 2 here -

https://github.com/google/EarlGrey/blob/master/docs/install-and-run.md#step-2-add-earlgrey-as-a-framework-dependency

You could change it to - 

> Because EarlGrey must be embedded within the app under test, we need to add certain Build Phases and Scheme changes to the Test Target. We recommend using our earlgrey gem to install any dependencies if you're using a CocoaPods version from `1.0.0` onwards. Run `gem install earlgrey`. You need to require this gem in a post_install hook using the project's name, the test target's name, and the name of the xcscheme file as is shown in the below example for `1.0.0`

You can move this passage to above the 0.39.0 example -

> Since the `earlgrey gem` is supported only beyond CocoaPods 1.0.0 or if you don't want a gem dependency, replace require 'earlgrey' with load 'configure_earlgrey.rb' and manually copy configure_earlgrey.rb from the gem into the same directory as the Podfile.

gurgeous:
How about this?

docs/install-and-run.md
```diff
@@ -64,6 +61,8 @@ For EarlGrey, we highly recommend [CocoaPods](https://cocoapods.org/) as the bes
   end
   ```
 
+    The earlgrey gem is supported with CocoaPods 1.0.0 and beyond. For earlier versions (or if you don't want a gem dependency) replace `require 'earlgrey'` with `load 'configure_earlgrey.rb'` and manually copy `configure_earlgrey.rb` from the gem into the same directory as the `Podfile`.
+
```
khandpur: This looks good but I'd also replace the actual code (in line # 79) with `load 'configure_earlgrey.rb'` for v0.39 and below. thanks!

gurgeous:
better?

bootstraponline:
Looks good. :+1:

khandpur:
Merging this. Thanks Adam!

Issue #221, opened 2016-08-15
configure_earlgrey - CommandLineArguments crash


gurgeous:
During `pod install` with cocoapods 1.0.1, see below. I believe this is because our xcscheme has `EnvironmentVariables` but not `CommandLineArguments`.

```
Generating Pods project
Checking and Updating HomeAway for EarlGrey.
Adding EarlGrey Framework Location as an Environment Variable
in the App Project's Test Target's Scheme Test Action.

[!] An error occurred while processing the post-install hook of the Podfile.

undefined method `elements' for nil:NilClass

/gems/2.1.0/gems/earlgrey-0.0.8/lib/earlgrey/configure_earlgrey.rb:251:in `add_environment_variables_to_test_action_scheme'
/gems/2.1.0/gems/earlgrey-0.0.8/lib/earlgrey/configure_earlgrey.rb:153:in `modify_scheme_for_actions'
/gems/2.1.0/gems/earlgrey-0.0.8/lib/earlgrey/configure_earlgrey.rb:112:in `configure_for_earlgrey'
/gems/2.1.0/gems/earlgrey-0.0.8/lib/earlgrey/configure_earlgrey.rb:18:in `configure_for_earlgrey'
Podfile:78:in `block (2 levels) in from_ruby'
```

bootstraponline:
Thanks for reporting this issue! It's a bug in the EarlGrey gem.

tirodkar:
@gurgeous - to confirm, you don't have <CommandLineArguments> present anywhere in your xcscheme file?

gurgeous:
My xcscheme file has `<EnvironmentVariables>` but no `<CommandLineArguments>`

tirodkar:
Our [demo app schemes](https://github.com/google/EarlGrey/tree/master/Demo/EarlGreyExample/EarlGreyExample.xcodeproj/xcshareddata) do not have any CommandLineArguments either. Would you mind posting your scheme or something analogous to it? This seems like a case we aren't covering.

gurgeous:
I cannot post my xcscheme, unfortunately. I diagnosed the bug by inspecting the code in the gem... Could be wrong, though. When I removed my `<EnvironmentVariables>` section I was able to complete the `pod install`.

bootstraponline:
It'd be nice if you could post a redacted version of the xcscheme so we can reproduce & resolve the bug. The schemes in our test suite contain `<EnvironmentVariables>` and they work correctly.

[...]

PR #223, opened 2016-08-17
Fix configure_earlgrey issue when either an Env Var or CommandLineArg is present in the Launch Action and not in the Test Action.

tirodkar:
Fixes https://github.com/google/EarlGrey/issues/221.

Issue - We have an OR check for if you have Env Variable or CommandLineArgs and your Test Action uses the vars from the launch action. However, if either one of them is empty, you can hit a nil element issue. This adds a check around that. Thanks @gurgeous

### DIFF
--- a/Demo/EarlGreyExample/EarlGreyExample.xcodeproj/xcshareddata/xcschemes/EarlGreyExampleTests.xcscheme
+++ b/Demo/EarlGreyExample/EarlGreyExample.xcodeproj/xcshareddata/xcschemes/EarlGreyExampleTests.xcscheme
@@ -36,6 +36,13 @@
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
       allowLocationSimulation = "YES">
+      <EnvironmentVariables>
+         <EnvironmentVariable
+            key = "FooKey"
+            value = "BarValue"
+            isEnabled = "YES">
+         </EnvironmentVariable>
+      </EnvironmentVariables>
       <AdditionalOptions>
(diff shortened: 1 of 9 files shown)


bootstraponline:
We need an automated test to detect the regression. I reverted `gem/lib/earlgrey/configure_earlgrey.rb` and even with the updated `EarlGreyExampleTests.xcscheme`, all the tests still pass. This is expected because the Ruby tests are based on fixtures. Updating the Demo project scheme does nothing.

tirodkar:
Agreed, I'll try and get the fixtures updated with a new project for this change.

tirodkar:
Updated. Added a couple of more projects.

bootstraponline:
I verified the tests detect the issue now. Looks good to me.

Demo/EarlGreyExample/EarlGreyExample.xcodeproj/xcshareddata/xcschemes/EarlGreyExampleTests.xcscheme
```diff
@@ -36,6 +36,13 @@
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
       allowLocationSimulation = "YES">
+      <EnvironmentVariables>
+         <EnvironmentVariable
+            key = "FooKey"
```
khandpur: So we know this is for testing purposes:

FooKeyForTesting
BarValueForTesting

khandpur:
LGTM

PR #232, opened 2016-08-19
Fix any breakages caused by the modification of the path to the gem lib

tirodkar:


docs/install-and-run.md
```diff
@@ -254,7 +254,7 @@ steps in your setup:
     #import <EarlGrey/EarlGrey.h>
    ```
 
-2. Download [`EarlGrey.swift`](https://github.com/google/EarlGrey/tree/master/gem/lib/earlgrey/files/EarlGrey.swift) and add it to your test bundle. The file contains
+2. Download [`EarlGrey.swift`](https://github.com/google/EarlGrey/tree/master/gem/lib/earlgrey/files/Swift-3.0/EarlGrey.swift) and add it to your test bundle. The file contains
```
khandpur: @bootstraponline 

we should be clear that this is flow is for manually including EarlGrey.swift outside of pods flow. Pods provides an automated way of doing that. Also, could we link the user to a directory where they can choose which version to get? I'd image when swift 3.1 or 4.0 comes out, we'll have to constantly update this...We could create a latest folder and symlink it to the latest and have our code and the default swift version in earlgrey gem reference that always?

bootstraponline: I think the easy solution is to link `https://github.com/google/EarlGrey/tree/master/gem/lib/earlgrey/files/` which will list all the Swift-* folders. Or move to a different folder structure: `files/swift/3.0` and link to `files/swift`

khandpur: `swift/3.0` is more organized but it requires updating the gem. Since this is a quick fix, I would suggest linking to `https://github.com/google/EarlGrey/tree/master/gem/lib/earlgrey/files/` first then in a follow-up PR make the dir structure, gem and documentation changes.

bootstraponline: Sounds good.

tirodkar:
Updated the `install_and_run.md` file.

tirodkar:
I've also added a commit that makes the EarlGreyExamples project's Swift tests to use the `EarlGrey.swift` file copied from the gem rather than directly from the directory path.

Demo/EarlGreyContribs/EarlGreyContribs.xcodeproj/project.pbxproj
```diff
@@ -105,6 +105,7 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
+		FD8277DB1D677D0D00C7C2AD /* EarlGrey.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; name = EarlGrey.swift; path = "../../../gem/lib/earlgrey/files/Swift-3.0/EarlGrey.swift"; sourceTree = "<group>"; };
```
khandpur: We should probably have a podfile for this too.

docs/install-and-run.md
```diff
@@ -254,7 +254,8 @@ steps in your setup:
     #import <EarlGrey/EarlGrey.h>
    ```
 
-2. Download [`EarlGrey.swift`](https://github.com/google/EarlGrey/tree/master/gem/lib/earlgrey/files/EarlGrey.swift) and add it to your test bundle. The file contains
+2. Download the correct [`EarlGrey.swift`](https://github.com/google/EarlGrey/tree/master/gem/lib/earlgrey/files/)
```
khandpur: But this only for non-pod flow. We need to make that clear otherwise you'd end up including earlgrey.swift twice.

Demo/EarlGreyContribs/EarlGreyContribs.xcodeproj/project.pbxproj
tirodkar: The Contribs project is specifically created to use EarlGrey from HEAD and not test the Pods flow out. Adding Pods would quite complicate this flow up.

tirodkar: Updated the doc.

docs/install-and-run.md
```diff
@@ -254,14 +254,27 @@ steps in your setup:
     #import <EarlGrey/EarlGrey.h>
    ```
 
-2. Download [`EarlGrey.swift`](https://github.com/google/EarlGrey/tree/master/gem/lib/earlgrey/files/EarlGrey.swift) and add it to your test bundle. The file contains
+2. In case you're using the [earlgrey gem](https://rubygems.org/gems/earlgrey), then you need to link to
```
khandpur: formatting is off (view the rendered md).

Organization wise, if you could bold the two ways of including earlgrey.swift, it would be clearer.

2. Include EarlGrey.swift in your project.

  **If using earlgrey gem:**
  `<command to include>`
  **If doing manual Installation:**
  `<manually link to the file>`

tirodkar: Added it.

docs/install-and-run.md
```diff
@@ -254,14 +254,34 @@ steps in your setup:
...
+      for your Swift version and add it to your test bundle. 
+
+    Here is an example of a very basic EarlGrey test written in Swift:
+
+    ```swift
+     // Select the button with Accessibility ID "clickMe" and assert it's visible.
+     EarlGrey().selectElementWithMatcher(grey_accessibilityID("ClickMe"))
+         .assertWithMatcher(grey_sufficientlyVisible())
+    ```
+3. For compatibility with Swift, we use `grey_allOfMatchers()` and `grey_anyOfMatchers()` instead of `grey_allOf()`
```
khandpur: Quick nit:

This seems like a helpful hint instead of a step a user must perform to enable swift support. Can we style it with ">" ?

> For compatibility with Swift, we use `grey_allOfMatchers()` and ...

docs/install-and-run.md
```diff
@@ -254,14 +254,34 @@ steps in your setup:
...
+
+  * **Manual installation without the [earlgrey gem](https://rubygems.org/gems/earlgrey)**
+
+      In this case you need to download the correct
+      [`EarlGrey.swift`](https://github.com/google/EarlGrey/tree/master/gem/lib/earlgrey/files/) file
+      for your Swift version and add it to your test bundle. 
+
+    Here is an example of a very basic EarlGrey test written in Swift:
+
+    ```swift
```
khandpur: another nit: This example isn't specific to manual copying of file. Currently it's indented with rest of content under 2.0, could we move it out?

Demo/EarlGreyContribs/EarlGreyContribs.xcodeproj/project.pbxproj
tirodkar: Updated.

khandpur:
LGTM from my end.

Issue #234, opened 2016-08-22
EarlGrey master broken on carthage


bootstraponline:
```
$ carthage version
0.17.2
```

```
$ carthage update EarlGrey --platform ios
*** Fetching EarlGrey
*** Checking out EarlGrey at "04ea8358bd607084ad52217da7db9142b9d48187"
*** xcodebuild output can be found in /var/folders/14/h2zk5sx1315b__blzdt7hz1m4k0by8/T/carthage-xcodebuild.cduOAU.log
*** Building scheme "EarlGrey" in EarlGrey.xcodeproj
** BUILD FAILED **


The following build commands failed:
       	CompileC /Library/Developer/Xcode/DerivedData/EarlGrey-higmssftwbthuodgjcrtayxetstw/Build/Intermediates/EarlGrey.build/Release-iphoneos/EarlGrey.build/Objects-normal/armv7/GREYAppStateTracker.o EarlGrey/Synchronization/GREYAppStateTracker.m normal armv7 objective-c com.apple.compilers.llvm.clang.1_0.compiler
(1 failure)
/eg_carthage_docs/Demo/EarlGreyExample/Carthage/Checkouts/EarlGrey/EarlGrey/Synchronization/GREYAppStateTracker.m:222:7: error: unused variable 'lock' [-Werror,-Wunused-variable]
/eg_carthage_docs/Demo/EarlGreyExample/Carthage/Checkouts/EarlGrey/EarlGrey/Synchronization/GREYAppStateTracker.m:225:7: error: unused variable 'unlock' [-Werror,-Wunused-variable]
A shell task (/usr/bin/xcrun xcodebuild -project /eg_carthage_docs/Demo/EarlGreyExample/Carthage/Checkouts/EarlGrey/EarlGrey.xcodeproj -scheme EarlGrey -configuration Release -sdk iphoneos ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean build) failed with exit code 65:
** BUILD FAILED **


The following build commands failed:
[...]

PR #235, opened 2016-08-23
Fix Carthage Break caused by Foundation Asserts being disabled in Release

tirodkar:
Fixes https://github.com/google/EarlGrey/issues/234


khandpur:
LGTM! Please wait for complete CI run before merging.

PR #243, opened 2016-08-31
Fix chunky_png ArgumentError

bootstraponline:
Work around https://github.com/applitools/Eyes.Selenium.Ruby/issues/42 by locking the dependencies to known good versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 .DS_Store
 xcuserdata
 *.gem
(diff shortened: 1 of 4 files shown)


khandpur:
LGTM from my end!

bootstraponline:
I updated the PR to use `eyes_selenium 2.37.0` which contains the fix for chunky_png.

PR #244, opened 2016-08-31
Updated cheatsheet and podspec for 1.2.0 release.

sid-github:


docs/cheatsheet/cheatsheet.html
```diff
@@ -31,13 +31,14 @@
 <div class="header">
 <div class="usage">
 <span class="one">EarlGrey()</span>.selectElementWithMatcher(<span class="two">ViewMatcher</span>)
+                     .indexOf(<span class="three">index</span>)
```
tirodkar: Um, I don't think this is needed. If so, we should be adding usingSearchAction etc. as well.

sid-github: Since cheatsheet is for quick reference I feel like we must add usingSearchAction too, what do you think?

tirodkar: Hmm, I'd prefer to keep this section as a reference on how to create a canonical GREYElementInteraction. This can get rather unwieldy if we add more API's here.

tirodkar:
LGTM. Thanks for this.

PR #252, opened 2016-09-07
Fix Ruby 2.2 on Travis

bootstraponline:
`osx_image: xcode7.3` means we're using [the `OS X 10.11` image](https://docs.travis-ci.com/user/osx-ci-environment/) which removed support for 2.2.2. Ruby 2.2.5 [is supported on 10.11.](http://rubies.travis-ci.org/)


khandpur:
thanks for unblocking our travis builds :)

PR #254, opened 2016-09-08
Move system alert constraint from GREYAction into GREYElementInteraction

axi0mX:
Move system alert view check into GREYElementInteraction so we don't have to check in every GREYAction.


tirodkar:
@axi0mX, since this change wasn't rebased recently and is currently conflicting, I'd suggest waiting a little since the GREYElementInteraction file has had quite a few modifications recently.

PR #256, opened 2016-09-08
make style more Swifty

KazuCocoa:
I think that:

- `.init` doesn't need in this case
- usage of `#file` and `#line` is suitable in this PR
    - https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/Expressions.html#//apple_ref/doc/uid/TP40014097-CH32-ID390

### DIFF
--- a/gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
+++ b/gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift
@@ -17,15 +17,15 @@
 import EarlGrey
 
 public func grey_allOfMatchers(args: AnyObject...) -> GREYMatcher! {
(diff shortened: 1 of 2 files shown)


tirodkar:
Thanks for this @KazuCocoa, can we also make this change for https://github.com/google/EarlGrey/tree/master/gem/lib/earlgrey/files/Swift-2.2/EarlGrey.swift ??

KazuCocoa:
Sure.
I just update this PR for Swift-2.2.

khandpur:
LGTM from my side. Thanks @KazuCocoa =]

KazuCocoa:
Thanks!

PR #286, opened 2016-09-21
Fix breaking Ruby tests by adding new spec projects.

tirodkar:
Fix ruby test breaks caused by the Example.xcodeproj in gem/fixture/specs not having a Products/ group within them. This is fixed by adding a more fleshed out project.


bootstraponline:
The Ruby build is still failing. I recommend disabling the `Style/VariableNumber` rubocop rule. It's a new rule they added. Following their recommendation results in less readable code.

tirodkar:
Let's not disable it. I'll fix the warnings. Thanks!

bootstraponline:
> Let's not disable it. I'll fix the warnings. Thanks!

Which is more readable:

- `swift22sha1`
- `swift_2_2_sha1`

Rubocop has some silly defaults which is why we're already disabling 8 rules.

tirodkar:
@bootstraponline Done. Thanks.

.rubocop.yml
```diff
@@ -35,3 +35,5 @@ Metrics/ModuleLength:
   Enabled: false
 Style/GuardClause:
   Enabled: false
+Style/VariableName:
+  Enabled:false
```
bootstraponline: `Enabled:false` should be `Enabled: false`

tirodkar:
Ruby tests look all green now.

Issue #287, opened 2016-09-21
configure_for_earlgrey fails if current directory doesn't contain the Podfile


ajsecord:
CocoaPods supports running `pod install` with a `--project_directory` option that should effectively act like

```sh
cd project_directory
pod install
```

When doing so with a CocoaPods project configured using the current EarlGrey installation instructions, using `--project_directory` causes `configure_for_earlgrey` to fail with the following error message:

```sh
Checking and Updating <MY_PROJECT> for EarlGrey.
[!] An error occurred while processing the post-install hook of the Podfile.

[Xcodeproj] Unable to open `<FULL_PATH_TO_WHERE_POD_INSTALL_WAS_RUN>/<MY_PROJECT>.xcodeproj` because it doesn't exist.

/Library/Ruby/Gems/2.0.0/gems/xcodeproj-1.3.1/lib/xcodeproj/project.rb:99:in `open'
/Library/Ruby/Gems/2.0.0/gems/earlgrey-0.0.13/lib/earlgrey/configure_earlgrey.rb:91:in `set_defaults'
/Library/Ruby/Gems/2.0.0/gems/earlgrey-0.0.13/lib/earlgrey/configure_earlgrey.rb:110:in `configure_for_earlgrey'
/Library/Ruby/Gems/2.0.0/gems/earlgrey-0.0.13/lib/earlgrey/configure_earlgrey.rb:18:in `configure_for_earlgrey'
/Users/ajsecord/Source/Git/material-components-ios2/catalog/Podfile:65:in `block (2 levels) in from_ruby'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-core-1.0.1/lib/cocoapods-core/podfile.rb:180:in `call'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-core-1.0.1/lib/cocoapods-core/podfile.rb:180:in `post_install!'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.1/lib/cocoapods/installer.rb:823:in `run_podfile_post_install_hook'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.1/lib/cocoapods/installer.rb:811:in `block in run_podfile_post_install_hooks'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.1/lib/cocoapods/user_interface.rb:141:in `message'
[...]

PR #294, opened 2016-10-03
Support project-directory in Pod Install call.

tirodkar:
fixes https://github.com/google/EarlGrey/issues/287.

@ajsecord, you'll have to wait for the next gem release (probably happening 4-5/10/2016) for getting this change with the gem. You should be able use `rake install` to get this change locally.

### DIFF
--- a/gem/lib/earlgrey/configure_earlgrey.rb
+++ b/gem/lib/earlgrey/configure_earlgrey.rb
@@ -20,19 +20,25 @@ end
 
 module EarlGrey
   class << self
-    attr_reader :project_name, :test_target, :test_target_name, :scheme_file,
-                :user_project, :carthage, :swift, :swift_version
+    attr_reader :project_name, :installer, :test_target, :test_target_name, :scheme_file,
+              :user_project, :carthage, :swift, :swift_version
 
     # Returns path to Xcode file, prepending current working dir if necessary.
-    # @param xcode_file [String] xcode file path
+    # @param project_name [String] name of the .xcodeproj file
     # @param ext [String] xcode file extension
     # @return [String] path to Xcode file
-    def path_for(xcode_file, ext)
-      ext_match = File.extname(xcode_file) == ext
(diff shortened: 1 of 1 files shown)


gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -20,21 +20,33 @@ def configure_for_earlgrey(*args)
...
+      path = File.join(dir_path, File.basename(project_name, '.*') + ext)
       path ? path : nil
     end
 
+    # Returns the project's directory. If CocoaPods hasn't had it passed in, then the current
+    # directory is chosen.
+    # @return [String] directory path for the Xcode project
+    def dir_path
+      if installer
+        pods_dir_path = installer.config.installation_root
```
bootstraponline: what is the value of `installer.config.installation_root` when `--project-directory` is not used?

gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -20,21 +20,33 @@ def configure_for_earlgrey(*args)
...
+        if pods_dir_path.to_s.length && Dir.pwd != pods_dir_path
```
bootstraponline: Why do we care if it's equal to Dir.pwd?

gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -20,21 +20,30 @@ def configure_for_earlgrey(*args)
...
     end
 
+    # Returns the project's directory. If CocoaPods hasn't had it passed in, then the current
+    # directory is chosen.
+    # @return [String] directory path for the Xcode project
+    def dir_path
+      if installer
+        path = installer.config.installation_root
+      end
+      path ? path : Dir.pwd
```
bootstraponline: why not just:

```ruby
def dir_path
  installer ? installer.config.installation_root : Dir.pwd
end
```

tirodkar: Done.

PR #298, opened 2016-10-07
Fix xcpretty dependency version to 0.2.2

tirodkar:


tirodkar:
https://github.com/google/EarlGrey/issues/302

PR #300, opened 2016-10-07
Update Info.plist / Podspec / Cheatsheet for EarlGrey 1.4.0

tirodkar:


docs/cheatsheet/cheatsheet.html
```diff
@@ -167,7 +167,7 @@
 
 <ul>
   <li>Assertions</li>
-  <li>Grey Assert
+  <li>UI Assertions
```
sid-github: Call this "Macros" instead?
GREYAsserts are not strictly for UI, we can use if for example to assert on timing in the test.

PR #306, opened 2016-10-11
Revert xcpretty to download the latest gem version

tirodkar:


tirodkar:
Updated!

khandpur:
We're back to having some strange issues. The [test is green but I see an unexpected failure reported in xcpretty](https://travis-ci.org/google/EarlGrey/jobs/166921933)

I checked rest of the links and found some discrepancies in the actual vs reported number of failures:
- [xcpretty claims *3 failed* but logs show only *2 failures*](https://travis-ci.org/google/EarlGrey/jobs/166921931)
- [xcpretty claims *3 failed* but logs show only *2 failures*](https://travis-ci.org/google/EarlGrey/jobs/166921932)

@supermarin We can file a bug on xcpretty but it may be hard to give you an exact repro because these failures are intermittent. Would it help if we passed you the entire xcodebuild.log file?

supermarin:
@khandpur I'd appreciate the raw log to peek in it.

This might be Xcode's own bug, as if you count the failures (x signs) themselves, there's really two.

khandpur:
@supermarin 
Logs are here: https://s3.amazonaws.com/archive.travis-ci.org/jobs/166921931/log.txt just below the text `cat xcodebuild.log`

It seems like there are indeed **3** failures:

1. Test Suite 'FTRAnalyticsTest' failed at 2016-10-12 01:35:52.343.
2. Test Case '-[FTRLocalUIWebViewTest testJavaScriptExecution]' failed (20.603 seconds).
3. Test Case '-[FTRLocalUIWebViewTest testSuccessiveTaps]' failed (41.687 seconds).

but **1.** fails in `+tearDown` so it's not tied to a test case per se.

tirodkar:
@supermarin, is there a specific place that xcpretty tries to get its failure details from?

supermarin:
@tirodkar there's patterns in `parser.rb`, although it's getting rewritten in the `blacklist` branch ATM. 
I'll take a closer look into this later today and ping back if it's something we can release easily in 0.2.x update.

khandpur:
I think we can this PR in and file a bug against xcpretty for the support for failures in +setUp/+tearDown

PR #307, opened 2016-10-13
Fixed failing analytics tests as we no longer sync with analytics requests

sid-github:
Fixed failing analytics tests as we no longer sync with analytics requests


sid-github:
FYI: I just updated this PR.

PR #311, opened 2016-10-17
Add iOS 10 Support to Travis

tirodkar:


.travis.yml
```diff
@@ -35,6 +44,8 @@ env:
     - DESTINATION="OS=8.1,name=iPhone 4S" SDK=iphonesimulator9.3 TYPE="FUNCTIONAL"
     - DESTINATION="OS=9.3,name=iPhone 6s" SDK=iphonesimulator9.3 TYPE="CARTHAGE"
```
khandpur: Could we group this line with similar destination and device?

bootstraponline:
The travis yml appears to fail lint...
http://lint.travis-ci.org/

tirodkar:
Fixed all the issues. This is good to merge now.

Scripts/travis.sh
```diff
@@ -61,12 +66,14 @@ execute_xcodebuild() {
...
 
 if [ "${TYPE}" == "RUBY" ]; then
-  rvm use 2.2.2;
+  rvm use 2.2.5;
   cd gem;
   bundle install --retry=3;
   rake;
 elif [ "${TYPE}" == "UNIT" ]; then
   execute_xcodebuild Tests/UnitTests/UnitTests.xcodeproj EarlGreyUnitTests
+elif [ "${TYPE}" == "FUNCTIONAL" && "${FILTER}" != "" ]; then
```
khandpur: Perhaps a better way would be:
Make the 3rd parameter of execute_xcodebuild be the filter (i.e. ${3}). Then in `execute_xcodebuild` method, check if ${3} is set, create a new variable assigned to`-only-testing ${3}` and pass it to `xcodebuild test` command.

tirodkar: After working with this for a bit, I reverted the script changes since as long as Xcode 8 doesn't support iOS 8 tests, we are not going to use the filter in our tests.

tirodkar:
This is good to go now. The script changes have been reverted since they're not being used anymore.

khandpur:
FYI, this surfaced another flakiness in unit tests:
https://travis-ci.org/google/EarlGrey/jobs/169643057

PR #312, opened 2016-10-19
Making it safe to invoke default failure handlers multiple times.

khandpur:
- It is safe to invoke failure handler multiple times without fearing loss of previous failure artifacts
- Adding some verbose logging and fixing minor typo [ci skip]


sid-github:
LGTM

PR #305, opened 2016-10-11
Fix all EarlGrey projects for Swift 3.0

tirodkar:
Fixes https://github.com/google/EarlGrey/issues/292.

This is only done to fix the extant EarlGrey projects. As you can see, the EarlGrey syntax seems quite off. For this purpose, we plan to add a newer version of EarlGrey.swift that suits the Swift nomenclature better and tries to remove any unneeded wrappers.

PR #314, opened 2016-10-19
More Swifty style for Swift 2.2(#file, #line, #selector)

sushichop:
**About `#file` and `#line`**
[Modernizing Swift's Debugging Identifiers](https://github.com/apple/swift-evolution/blob/master/proposals/0028-modernizing-debug-identifiers.md)

**About `#selector`**
[Referencing the Objective-C selector of a method](https://github.com/apple/swift-evolution/blob/master/proposals/0022-objc-selectors.md)

These styles have been introduced from Xcode 7.3 with Swift 2.2.


tirodkar:
@sushichop Thanks for the change! Quick heads up: this code will be moved / overwritten in https://github.com/google/EarlGrey/pull/305.

PR #315, opened 2016-10-19
Fix travis builds to fail on any unexpected failures

khandpur:
unexpected failures in +setUp or +tearDown don't count as test failures so xcode doesn't return a failure code upon exit. XCPretty also overlooks unexpected failures that arise in +setUp and +tearDown methods. The fix is to grep for "0 unexpected failures" in the final output and fail the test run if we don't find it.

### DIFF
--- a/Scripts/travis.sh
+++ b/Scripts/travis.sh
@@ -37,26 +37,41 @@ execute_xcodebuild() {
     exit 1
   fi
 
-  retval_xcodebuild=0
+  retval_test_command=0
   for retry_attempts in {1..3}; do
-    # To retry on failure, disable exiting if command below fails.
+    # As we are attempting retries, disable exiting when command below fails.
     set +e
     env NSUnbufferedIO=YES xcodebuild -project ${1} -scheme ${2} -sdk "$SDK" -destination "$DESTINATION" -configuration "$CONFIG" ONLY_ACTIVE_ARCH=NO $ACTION | tee xcodebuild.log | xcpretty -sc;
-    retval_xcodebuild=$?
-    # Even failed tests exit with code 65, check to ensure tests haven't started.
+    retval_test_command=$?
+
+    # Retry condition 1: Tests haven't started.
     # We achieve that by looking for keyword "Test Suite" in xcodebuild.log.
-    # TODO: this behavior may change.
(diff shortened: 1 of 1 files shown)


Scripts/travis.sh
```diff
@@ -37,26 +37,48 @@ execute_xcodebuild() {
     exit 1
   fi
 
-  retval_xcodebuild=0
+  retval_command=0
```
sid-github: use local to mark retval_command as a local var.

Scripts/travis.sh
```diff
@@ -27,7 +27,7 @@ ACTION="test"
 #  $1 : .xcodeproj file
 #  $2 : scheme to run
 #
-# The output is prettified using xcpretty and redirected to xcodebuild.log for failure analysis.
+# The output is formatted using xcpretty and redirected to xcodebuild.log for failure analysis.
 execute_xcodebuild() {
   if [ -z ${1+x} ]; then
```
sid-github: Remove x (also below):
if [ -z ${1} ]; then
is what you need, this probably worked because unset x is being treated as 0.

Scripts/travis.sh
```diff
@@ -37,26 +37,48 @@ execute_xcodebuild() {
     exit 1
   fi
 
-  retval_xcodebuild=0
+  retval_command=0
+  # Are we running a test?
+  [[ "${ACTION}" == *"test"* ]] && is_running_test=1 || is_running_test=0
```
sid-github: nit: this oneliner is neat! However for readability sake consider changing this to an if else statement.

Scripts/travis.sh
khandpur: No, I don't think that's the right reason you pointed out. Here's a more detailed explanation: http://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash

Scripts/travis.sh
khandpur: if / else just bloats code here; I think this shorthand way fine.

khandpur:
Here's an example of it catching the unexpected failure:
https://travis-ci.org/google/EarlGrey/jobs/169423327

tirodkar:
Squashing and merging this in since https://travis-ci.org/google/EarlGrey/jobs/169423333 shows the improvement being made here.

PR #329, opened 2016-11-01
Allow Clear Text to accept Objects that implement the UITextInput protocol even though they do not respond to the @text selector.

tirodkar:


sid-github:
LGTM

PR #338, opened 2016-11-10
Using bit-dependent trig functions to deal with different underlying …

khandpur:
…types of CGFloat on 32 and 64 bit


tirodkar:
LGTM

PR #352, opened 2016-11-22
Fix EarlGrey CocoaPods project and add a Travis target for it. 

tirodkar:


Demo/EarlGreyExample/Podfile
```diff
@@ -41,4 +49,4 @@ post_install do |installer|
     end
     configure_for_earlgrey(installer, PROJECT_NAME, test_target, test_target + '.xcscheme', var_args)
   end
-end
+end
```
khandpur: no new line?

tirodkar:
I updated this PR. We also fix some comments in the gem and add a way to run the CocoaPods project with the local gem from travis.

tirodkar:
The build is fixed here. I'm not providing a file path for the `rake install`but instead entering it and installing the gem.

sid-github:
LGTM

PR #384, opened 2017-01-05
Swift 3.0 fixes

bootstraponline:
Swift 3 fixes from Xcode 8.2.1 migrator.

### DIFF
--- a/gem/lib/earlgrey/files/Swift-3.0/EarlGrey.swift
+++ b/gem/lib/earlgrey/files/Swift-3.0/EarlGrey.swift
@@ -60,13 +60,13 @@ public func GREYAssertNotEqualObjects<T: Equatable>( _ left: @autoclosure () -> 
 }
 
 public func GREYFail(_ reason: String) {
-  EarlGrey.handle(exception: GREYFrameworkException(name: kGREYAssertionFailedException,
+  EarlGrey.handle(GREYFrameworkException(name: kGREYAssertionFailedException,
                                                     reason: reason),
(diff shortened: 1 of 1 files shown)


tirodkar:
LGTM. Waiting on the CI.

tirodkar:
QQ - will this work in older Xcode versions? What if a user has Xcode 8.0 or 8.1?

bootstraponline:
> QQ - will this work in older Xcode versions? What if a user has Xcode 8.0 or 8.1?

I think it will break them. I don't know for sure though.

tirodkar:
Cool. Let's wait for the CI to run and see the result. In case there's a failure, we might have to move this to a new EarlGrey.swift version and make changes to the Travis Xcode versions for CI checks, which currently does not have 8.2.1 support.

khandpur:
Travis Swift test fails: https://travis-ci.org/google/EarlGrey/jobs/189054172

bootstraponline:
How do we want to version this? It looks like Xcode 8.2.1 is still considered `Swift 3`

bootstraponline:
I reverted the label changes. I expect Travis to pass now.

bootstraponline:
Thoughts on moving all the `public` access to `open`? Right now I just have the ones that Xcode complained about.

- https://github.com/apple/swift-evolution/blob/master/proposals/0117-non-public-subclassable-by-default.md

tirodkar:
We should move them all to open as per the swift-evolution suggestion.

PR #391, opened 2017-01-12
Moves autorelease pool to the intended position.

stkhapugin:
The autorelease pool clearly was intended to be used inside the loop. Fixing this typo so it actually releases objects in a tight loop. 

We found this bug as part of investigation of https://bugs.chromium.org/p/chromium/issues/detail?id=675399#c9


khandpur:
Just reran the swift targets on travis as they error-ed out due to no output (unrelated to this fix).

stkhapugin:
Thanks for the quick review!

PR #394, opened 2017-01-19
Fix to compute closest inside integral rect

wuhao5:


wuhao5:
the travis failure doesn't seem to relate to this change, but will re-run and take another look at why.

wuhao5:
I have re-run the travis and the errors are different, I think those are just flaky. also the ruby failure seems to be related to the travis environment setup, I wonder if we should fix that in a different PR (so the commit history is cleaner and easier to do the release merge).

@khandpur @tirodkar @sid-github let me know how do you want me to proceed

khandpur:
yes, lets address them in a different PR

PR #397, opened 2017-01-19
Update EarlGrey.podspec.json

kebernet:
Updating zip URL to use HTTPS.


tirodkar:
Could you sign the CLA please? Also, what is the advantage of using `https`?

kebernet:
I signed it!

I don't like the idea of DNS spoofing potentially introducing unknown code into our app, to be honest.

PR #399, opened 2017-01-19
disable blockline limit

wuhao5:
attempt to fix this error:

https://travis-ci.org/google/EarlGrey/jobs/193486446

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,6 +14,8 @@
 #
 Metrics/LineLength:
   Max: 100
+Metrics/BlockLength:
+  Enabled: false
 Style/PercentLiteralDelimiters:
   PreferredDelimiters:
     '%i': '[]'


wuhao5:
waiting on the travis

wuhao5:
passed here, https://travis-ci.org/google/EarlGrey/jobs/193541262

khandpur:
@tirodkar @bootstraponline  PTAL

bootstraponline:
I agree with the title `disable blockline limit` however that's not what this PR does.

Instead `.rubocop.yml` should be modified:

```
Metrics/BlockLength:
  Enabled: false
```

A bunch of other metrics are already ignored because they don't provide value. The Ruby code shouldn't be modified to conform to this.

khandpur:
That's what the PR did originally but I asked if we could make ruby code conform to the limit. @bootstraponline - Is that not the standard? We don't have a proper ruby style-guide so taking rubocop enforced style as gospel.

bootstraponline:
> @bootstraponline - Is that not the standard?

Nope, it's a random number that someone picked. Modifying the code to conform to it makes the tests less readable.

> We don't have a proper ruby style-guide so taking rubocop enforced style as gospel.

Almost all Ruby projects use a customized version of the rubocop styles. It's not like golang where there's a single standard that everyone follows. Most of the rubocop styles are great. It's specifically the `Metrics/*` ones that are annoying.

bootstraponline:
FYI even rubocop itself ignores blocklength for specs:

- https://github.com/bbatsov/rubocop/blob/157b947473e32e6e79bedfa57798fdf9fc6a0650/.rubocop.yml#L19

khandpur:
Thanks @bootstraponline I wasn't aware. @wuhao5 please revert back to your original fix.

wuhao5:
Done, waiting on travis here: https://travis-ci.org/google/EarlGrey/jobs/193826120

Thanks @bootstraponline and @khandpur

PR #408, opened 2017-01-21
Using custom text matcher instead of hasProperty because hasProperty …

wuhao5:
…formats the description differently and it looks bad with the new error message logs.

Github import of changes:

  - Using custom text matcher instead of hasProperty because hasProperty formats the description differently and it looks bad with the new error message logs.

Old one shown as: an object with text "l0rem.ipsum+42@google.com#*-")
New one shown as: hasText('l0rem.ipsum+42@google.com#*-')

PiperOrigin-RevId: 145150301


khandpur:
Droping this in favor of #409

PR #401, opened 2017-01-20
Update error logging and reporting format

wuhao5:
https://github.com/google/EarlGrey/issues/392

@phsong-google @khandpur @sid-github @tirodkar

PR #409, opened 2017-01-21
Using custom text matcher instead of hasProperty matcher.

khandpur:
Github import of changes:

  - Using custom text matcher instead of hasProperty because hasProperty formats the description differently and it looks bad with the new error message logs.

Old one shown as: an object with text "l0rem.ipsum+42@google.com#*-")
New one shown as: hasText('l0rem.ipsum+42@google.com#*-')

PiperOrigin-RevId: 145150301


wuhao5:
I think this one needs to wait for https://github.com/google/EarlGrey/pull/401, but both of those are good to merge

PR #413, opened 2017-01-25
Improving error reporting for failures caused by thread executor timeouts.

wuhao5:
Github import of changes:

  - - Removing JSON escape in reported errors. It makes it confusing to understand the error by escaping string values.
- Presenting ThreadExecutor timeouts in a better format:

PiperOrigin-RevId: 145506353


wuhao5:
the CI looks all good, ready to merge

PR #423, opened 2017-02-04
Add test to long press a UITextField.

mbaxley:
This often causes the app state tracker to never become idle
on subsequent tests on iOS 9.3. It happend occasionally on simulator
and virtually every time on device.


tirodkar:
If this is a failing test case, could we have it marked DISABLED_ and we'll enable it when providing the fix?

tirodkar:
Also, any particular device this is failing on?

mbaxley:
I repro'd it about 3 out of 10 times on an iPad Air 2 simulator (iOS 9.3). The device where I could repro it ~100% of the time was an iPad Mini, running iOS 9.3.5.

I have updated the PR to disable the test.

PR #431, opened 2017-02-16
Blacklist UICompatibilityInputViewController from being tracked in UIViewController tracking.

tirodkar:
A user finds that tracking it causes an accessory view to not be untracked on a keyboard, thus halting any test that types with it.

### DIFF
--- a/EarlGrey/Additions/UIViewController+GREYAdditions.m
+++ b/EarlGrey/Additions/UIViewController+GREYAdditions.m
@@ -23,10 +23,17 @@
 #import "Common/GREYSwizzler.h"
 #import "Synchronization/GREYAppStateTracker.h"
 
+/**
+ *  The class for UICompatibilityInputViewController which isn't tracked here since we've faced
+ *  issues with tracking it when it comes to typing on keyboards with accessory views.
+ */
+static Class compatibilityVCClass;
+
 @implementation UIViewController (GREYAdditions)
 
 + (void)load {
   @autoreleasepool {
+    compatibilityVCClass = NSClassFromString(@"UICompatibilityInputViewController");
     GREYSwizzler *swizzler = [[GREYSwizzler alloc] init];
     // Swizzle viewWillAppear.
     BOOL swizzleSuccess = [swizzler swizzleClass:self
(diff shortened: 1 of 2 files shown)


khandpur:
@sid-github please merge if code + ci lgty.

tirodkar:
There was a small failure in the 4s tests. Fixed it.

heemoe:
is the latest release including this PR?

tirodkar:
@heemoe Yes 1.7.2 contains this change.

heemoe:
@tirodkar I'm just updated latest version , but my project still have the similar error.

some log here:
```
<_UITableViewCellBadgeNeue:0x7fd5526404a0> => Waiting for UIView's draw/layout pass to complete. A draw/layout pass normally completes in the next runloop drain.
```

```
<UILabel:0x7fd552437600> => Waiting for UIView's draw/layout pass to complete. A draw/layout pass normally completes in the next runloop drain.
```

```
<UIInputWindowController:0x7fd54e1aca00> => Waiting for viewDidAppear: call on this view controller. Please ensure that this view controller and its subclasses call through to their super's implementation
Waiting for window's rootViewController to appear. This should happen in the next runloop drain after a window's state is changed to visible.
```

tirodkar:
@heemoe Could you provide us details on how your test is? I have a sample test [here](https://github.com/google/EarlGrey/blob/master/Tests/FunctionalTests/Sources/FTRKeyboardKeysTest.m#L460) that ignores the Input Accessory and types in a UITextField which this PR enabled. Any way you could repro your issue with our project or so?

heemoe:
@tirodkar I got these issues even do nothing in my test case. just import header.

tirodkar:
I see. That makes sense. In case you're creating your own accessory or in the view controller, could you check if you might be missing a `[super viewDidLoad]` or so?

heemoe:
@tirodkar  This is a swift project. I did't found any the function is missing  super.viewDidLoad or so. Even I new a test file , these issues also happened.

```
//
//  newTest.swift
//
//  Created by Hee on 20/02/2017.
//

import XCTest

class newTest: XCTestCase {
    
    override func setUp() {
        super.setUp()
        // Put setup code here. This method is called before the invocation of each test method in the class.
    }
    
    override func tearDown() {
        // Put teardown code here. This method is called after the invocation of each test method in the class.
        super.tearDown()
    }
    
    func testExample() {
        // This is an example of a functional test case.
        // Use XCTAssert and related functions to verify your tests produce the correct results.
    }
    
    func testPerformanceExample() {
        // This is an example of a performance test case.
        self.measure {
            // Put the code you want to measure the time of here.
        }
    }
    
}

```

tirodkar:
That's very odd. If it was something only EarlGrey was doing then our Swift builds should all have failed as well. Are you able to repro this with any of our test projects as well?

We've added a check for UICompatibilityInputViewController in this PR. Your issue might be another case we're not looking for. Is there any specific way that you're adding the input accessory?

heemoe:
There is a framework that using to add input accessory. I tried to remove the framework and clean cache , rebuild , but it does't work. I think it not just about input accessory.  Should I create a new issue about this?

tirodkar:
@heemoe that would be fine. It would be great if you could also provide repro steps for us.

heemoe:
@tirodkar  It's hard to repro. I have no idea.

PR #437, opened 2017-02-22
Fix breaks in the EarlGrey Contribs tests caused by the fact that the Basic Views Cell is obscured in iPhone 4s

tirodkar:


Demo/EarlGreyContribs/EarlGreyContribsSwiftTests/EarlGreyContribsSwiftTests.swift
```diff
@@ -27,6 +27,8 @@ class EarlGreyContribsSwiftTests: XCTestCase {
 
   func testBasicViewController() {
     EarlGrey().selectElementWithMatcher(grey_text("Basic Views"))
+      .usingSearchAction(grey_scrollInDirection(.Down, 50),
+        onElementWithMatcher: grey_kindOfClass(UITableView.self))
```
khandpur: is the indentation correct?

tirodkar: 2 -> 2 spaces. I believe this is correct.

PR #451, opened 2017-03-16
Enable accessibility on device on iOS 9.1 +

wuhao5:
On iOS 9.1 and higher, -[XCAXClient_iOS loadAccessibility:] is needed, but it is
not available on older versions of XCTest/Xcode, replacing with
-[respondToSelector:] check.


wuhao5:
the failures doesn't relate to this change, it should be good to go!

PR #455, opened 2017-03-17
Use same copy files script for both cocoapods & carthage

### LINKED ISSUE #454: Replace carthage copy-frameworks with regular copy script
Carthage's copy-frameworks doesn't actually copy the frameworks. Instead we can reuse the existing cocoapod's copy frameworks script.

carthage_copy_files_fail

https://github.com/google/EarlGrey/blob/master/docs/install-and-run.md#final-test-configuration-

bootstraponline:
Fix #454

### DIFF
--- a/gem/lib/earlgrey/configure_earlgrey.rb
+++ b/gem/lib/earlgrey/configure_earlgrey.rb
@@ -120,8 +120,7 @@ module EarlGrey
       scheme = modify_scheme_for_actions
 
       # Add a Copy Files Build Phase for EarlGrey.framework to embed it into the app under test.
-      # carthage uses carthage copy-frameworks instead of a copy files build phase.
-      add_earlgrey_copy_files_script unless carthage
+      add_earlgrey_copy_files_script
 
       # Add header/framework search paths for carthage
       add_carthage_search_paths
@@ -384,17 +383,6 @@ module EarlGrey
       unless existing_frameworks.include? earlgrey_framework
         framework_ref = add_earlgrey_product
         test_target.frameworks_build_phase.add_file_reference framework_ref
-      end
-
-      if carthage
-        # Add shell script phase
-        shell_script_name = 'Carthage copy-frameworks Run Script'
-        unless test_target.shell_script_build_phases.map(&:name).include?(shell_script_name)
-          shell_script = test_target.new_shell_script_build_phase shell_script_name
-          shell_script.shell_path = '/bin/bash'
-          shell_script.shell_script = '/usr/local/bin/carthage copy-frameworks'
-          shell_script.input_paths = ['$(SRCROOT)/Carthage/Build/iOS/EarlGrey.framework']
-        end
       end
 
       user_project.save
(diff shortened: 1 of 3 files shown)


gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -120,8 +120,7 @@ def configure_for_earlgrey(installer, project_name, test_target_name, scheme_fil
       scheme = modify_scheme_for_actions
 
       # Add a Copy Files Build Phase for EarlGrey.framework to embed it into the app under test.
-      # carthage uses carthage copy-frameworks instead of a copy files build phase.
-      add_earlgrey_copy_files_script unless carthage
+      add_earlgrey_copy_files_script
```
khandpur: nice

gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -386,17 +385,6 @@ def copy_swift_files
         test_target.frameworks_build_phase.add_file_reference framework_ref
       end
 
-      if carthage
```
khandpur: double nice

gem/spec/fixtures/carthage_after/Example.xcodeproj/project.pbxproj
```diff
@@ -7,10 +7,11 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		5B5678F4326B273D6B2DD1F4 /* EarlGrey.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 0B657D1D6749CAC16B3688D5 /* EarlGrey.framework */; };
+		7D3DE2B2D4B6ADBAD8023E1F /* EarlGrey.swift in Sources */ = {isa = PBXBuildFile; fileRef = 23083A8BBED0E8BDFEA08B61 /* EarlGrey.swift */; };
+		9F0F538927F2B2C5677F9D15 /* EarlGrey.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E96DE017C569DE30A62E26B5 /* EarlGrey.framework */; };
```
khandpur: why is EarlGrey.framework appearing twice here.

bootstraponline: One is for frameworks and the other is for the copy files script. Cocoapods is setup the same way if you look at that fixture.

```ruby
user_project.products_group.new_file(framework_path)

test_target.frameworks_build_phase.add_file_reference framework_ref
```

are you saying it'd be better to cache the framework ref and use the same one in both locations?

bootstraponline: In `add_earlgrey_product` I'm caching the file reference to `EarlGrey.framework` and I just added `avoid_duplicates: true` to the `add_file_reference` calls. It's still ending up with two file references.

bootstraponline: I think this is why we get two duplicates:

```
build_file = new_copy_files_phase.add_file_reference(file_ref, true)
build_file.settings = { 'ATTRIBUTES' => ['CodeSignOnCopy'] }
```

Adding the attribute creates a different file ref identifier.

PR #457, opened 2017-03-17
Fix grey_getText

### LINKED ISSUE #456: Capturing the text value of an element
This function in the FAQ for getting the string value of aN element no longer works in Swift 3. When trying to use the method, you encounter to small issues before hitting the real one
1. `inout`'s are not composed in the example the way they are in the example in Swift 3 
2. The method itself has changed in EarlGrey (small change, the compiler takes care of that for you)
3. After the fixing those minor issues you hit an error **"escaping closure can only capture inout parameters explicitly by value"**.  This is due to a swift 3 change and seems reasonable given the explanation [here](https://github.com/apple/swift-evolution/blob/master/proposals/0035-limit-inout-capture.md) 

From the [EarlGrey FAQ](https://github.com/google/EarlGrey/blob/master/docs/faq.md)
> The same technique works for extracting element attributes into variables. Here's an example of storing an element's text attribute.
> 
>    Example Usage:
> ```
>    var textValue:String = "" 
>    domainField.performAction(grey_typeText("hello.there"))
>        .performAction(grey_getText(&textValue))
> 
>  
> func grey_getText(inout text: String) -> GREYActionBlock {
>   return GREYActionBlock.actionWithName("get text",
>     constraints: grey_respondsToSelector(Selector("text")),
>     performBlock: { element, errorOrNil -> Bool in
>       text = element.text
>       return true
>     })
> }
> ```
> 

**Given one can't use inouts the same way in Swift 3, how can the text value of an element be retrieved?**

bootstraponline:
Fix #456

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -75,7 +75,8 @@ open class Element {
  *  GREYAssertTrue(element.text != "", reason: "get text failed")
  */
 public func grey_getText(_ elementCopy: Element) -> GREYActionBlock {
-  return GREYActionBlock.action(withName: "get text", constraints: grey_respondsToSelector(#selector(getter: UILabel.text))) { element, errorOrNil -> Bool in
+  return GREYActionBlock.action(withName: "get text", 
+  constraints: grey_respondsToSelector(#selector(getter: UILabel.text))) { element, errorOrNil -> Bool in
         let elementObject = element as? NSObject
         let text = elementObject?.perform(#selector(getter: UILabel.text), with: nil)?.takeRetainedValue() as? String
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -57,21 +57,31 @@ Here's an example of storing an element's text attribute.
 
 ```swift
 // Swift
-/**
+//
+// Must use wrapper class to force pass by reference in block.
+// inout params won't work. http://stackoverflow.com/a/28252105
+open class Element {
+  var text = ""
+}
+
+/*
  *  Example Usage:
  *
- *  var textValue:String = ""
+ *  let element = Element()
  *
- *  domainField.performAction(grey_typeText("hello.there"))
- *      .performAction(grey_getText(&textValue))
+ *  domainField.performAction(grey_replaceText("hello.there"))
+ *             .performAction(grey_getText(element))
+ *
+ *  GREYAssertTrue(element.text != "", reason: "get text failed")
  */
-func grey_getText(inout text: String) -> GREYActionBlock {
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -57,21 +57,31 @@ Here's an example of storing an element's text attribute.
...
-  return GREYActionBlock.actionWithName("get text",
-    constraints: grey_respondsToSelector(Selector("text")),
-    performBlock: { element, errorOrNil -> Bool in
-      text = element.text
-      return true
-    })
+public func grey_getText(_ elementCopy: Element) -> GREYActionBlock {
+  return GREYActionBlock.action(withName: "get text", constraints: grey_respondsToSelector(#selector(getter: UILabel.text))) { element, errorOrNil -> Bool in
+        let elementObject = element as? NSObject
+        let text = elementObject?.perform(#selector(getter: UILabel.text), with: nil)?.takeRetainedValue() as? String
```
tirodkar: Assuming the users know what they're looking for, let's just change this to be on a UILabel:
```
    return GREYActionBlock.action(withName: "get text",
                                   constraints: getTextContraint,
                                   perform: { (element, errorOrNil) -> Bool in
                                    let label = element as! UILabel
                                    contain = label.text!
                                    return true
    })
```

docs/faq.md
```diff
@@ -57,21 +57,31 @@ Here's an example of storing an element's text attribute.
...
  */
-func grey_getText(inout text: String) -> GREYActionBlock {
-  return GREYActionBlock.actionWithName("get text",
-    constraints: grey_respondsToSelector(Selector("text")),
-    performBlock: { element, errorOrNil -> Bool in
-      text = element.text
-      return true
-    })
+public func grey_getText(_ elementCopy: Element) -> GREYActionBlock {
+  return GREYActionBlock.action(withName: "get text", constraints: grey_respondsToSelector(#selector(getter: UILabel.text))) { element, errorOrNil -> Bool in
```
khandpur: nit: can we break so that `constraints: grey_respondsToSelector(#selector(getter: ...` is in it's own line? will help readability.

docs/faq.md
bootstraponline: per https://github.com/realm/SwiftLint/pull/693 I'd like to avoid force casts / force unwraps

bootstraponline:
sure, I updated the text. the PR is set to allow edits from maintainers. feel free to make any adjustments as necessary. 

I will be back on Monday. 😄

Issue #466, opened 2017-03-21
Improve earlgrey gem to set up the project easily


wuhao5:
Proposing the following improvements in the earlgrey gem:

1. remove post_install, earlgrey.pod would automatically set up the target:
```ruby
target "ExampleTests" do
  pod 'EarlGrey'
end

# remove post_install
```
This will add the copy script and modify the schemes for the target "ExampleTests" only when it is a unit test bundle.

2. config xcscheme files per target
It would apply required changes to all the schemes that include the given target; currently, it targets only the given test target; sometimes, the app target can contain the test target as well.

3. copy EarlGrey.swift when the test target has a swift source code or `swift_version` is defined in podfile:
```ruby
target "ExampleTests" do
  pod 'EarlGrey'
  swift_version = '3.0'
end
```
Note if the project has `SWIFT_VERSION` or `Use Legacy Swift Language Version` defined, it will also include the swift file

tirodkar:
How would the `the app target can contain the test target`? As in we have a target that consolidates multiple targets? We should make it clear not to have that.

PR #467, opened 2017-03-22
Improve earlgrey gem #466

wuhao5:
1. remove post_install, the pod itself will update the proj
2. config schemes and swifts dynamically


wuhao5:
https://github.com/google/EarlGrey/issues/466

to test it out, you can do:
```
cd gem && rake install:local  # showing earlgrey 0.2 is installed
cd Demo/EarlGreyExample/ && pod install
```

I did monkey-patch for `Analyzer#generate_target` and `Analyzer#generate_targets`. This may be a problem because those two methods are internally used - it can change and break here. `Analyzer#analyze` seems to be a better option, will look into that next.

Also I'd like hear feedbacks for podfile extension, which would require the user to add `use_earlgrey!` in the podfile and it's even more intrusive.

gem/Gemfile
```diff
@@ -16,8 +16,3 @@
 source 'https://rubygems.org'
 
 gemspec
-
-gem 'coveralls', require: false
-gem 'rspec'
-gem 'rubocop'
-gem 'thor' # enable: bundle exec thor spec
```
bootstraponline: why did you delete the gems from the Gemfile? 😭 This broke the bundle spec command.

```
$ bundle exec thor spec
Could not find gem 'coveralls (~> 0.8.13)' in any of the gem sources listed in your Gemfile.
Run `bundle install` to install missing gems.
```

gem/lib/cocoapods_plugin.rb
```diff
@@ -0,0 +1,16 @@
+#
+#  Copyright 2016 Google Inc.
```
bootstraponline: 2016?

gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -13,19 +13,44 @@
...
+
+def configure_for_earlgrey
+  raise 'configure_for_earlgrey is deprecated, please remove.'
 end
 
 module EarlGrey
+  XCScheme = Xcodeproj::XCScheme
+  EnvironmentVariable = XCScheme::EnvironmentVariable
+
+  XCSCHEME_EXT = '*.xcscheme'.freeze
```
bootstraponline: in modern ruby, strings are frozen by adding a magic comment to the top of the file

`# frozen_string_literal: true`

rubocop will even warn you if this comment isn't present

it's preferable to use this approach instead of invoking .freeze manually on every string.

gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -108,243 +128,231 @@ def set_defaults(installer, project_name, test_target_name, scheme_file, opts =
...
+      modify_scheme_for_actions(user_project, [test_target]).each do |_, scheme|
+        scheme.save!
+      end
 
-      # Add a Copy Files Build Phase for EarlGrey.framework to embed it into the app under test.
-      add_earlgrey_copy_files_script
+      # Add a Copy Files Build Phase for EarlGrey.framework to embed it into
+      # the app under test. Carthage uses carthage copy-frameworks instead of
+      # a copy files build phase.
+      framework_ref = add_earlgrey_product user_project, carthage
```
bootstraponline:
```ruby
# Carthage uses carthage copy-frameworks instead of
# a copy files build phase.
```
^ that should be deleted, we're now using the same `Copy Files Build Phase` for both cocoapods and carthage.

gem/spec/spec_helper.rb
```diff
@@ -13,7 +13,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-require_relative '../lib/earlgrey'
+require 'earlgrey/cli'
```
bootstraponline: the spec should run using earlgrey from source code, not the earlgrey installed gem.

required_relative is how we ensure we're testing the actual earlgrey code.

bootstraponline:
do the carthage steps [as defined here](https://github.com/google/EarlGrey/tree/master/Demo/EarlGreyExample#getting-started---carthage) work for you on the Example project using the new gem changes?

I keep getting:

> Test target EarlGreyExampleSwiftTests encountered an error (Early unexpected exit, operation never finished bootstrapping - no restart will be attempted)

happens regardless of the gem version though.

khandpur:
Confirming that after installing the earlgrey gem which includes changes in this PR, the scheme, copy files script and earlgrey framework were correctly setup for the demo project.

bootstraponline:
@khandpur did the tests (`EarlGreyExampleSwiftTests`) run successfully though? I agree the setup looks good.

gem/spec/spec_helper.rb
wuhao5: I couldn't find rspec doc but here is a [SO link](http://stackoverflow.com/questions/21544519/correct-way-to-test-lib-folder-with-rspec):
rspec should set up `$LOAD_PATH`

bootstraponline: `require 'earlgrey/cli'` can load from gems `require_relative` can't

we don't want the spec loading from the earlgrey installed gem. I don't think we have any guarantees what the $LOAD_PATH looks like.

khandpur:
@bootstraponline yup both `EarlGreyExampleSwiftTests` and `EarlGreyExampleTests` ran fine for me. Did they fail for you?

bootstraponline:
> @bootstraponline yup both EarlGreyExampleSwiftTests and EarlGreyExampleTests ran fine for me.

awesome.

> Did they fail for you?

I think it's some local issue with my Xcode, the demo project isn't working for me (`operation never finished bootstrapping`). It seems fully unrelated to this PR though. It was working fine a few days ago. ❤️ 🍎

gem/lib/earlgrey/configure_earlgrey.rb
wuhao5: good to know, thanks! looks like rubocop will warn `.freeze` on prior ruby2.3, I wanted to at least make sure this works with the built-in ruby as most users won't upgrade their Ruby just for installing a gem.

gem/Gemfile
wuhao5: it should be covered from `gemspec` in the line above. what did it give you when you do `bundle list`, here is mine:

```
Gems included by the bundle:
  * CFPropertyList (2.3.5)
  * activesupport (4.2.8)
  * ast (2.3.0)
  * bundler (1.14.6)
  * claide (1.0.1)
  * coderay (1.1.1)
  * colored (1.2)
  * coveralls (0.8.19)
  * diff-lcs (1.3)
  * docile (1.1.5)
  * earlgrey (0.2.0)
  * i18n (0.8.1)
  * json (2.0.3)
  * method_source (0.8.2)
  * minitest (5.10.1)
  * nanaimo (0.2.3)
  * parser (2.4.0.0)
  * powerpack (0.1.1)
  * pry (0.10.4)
  * rainbow (2.2.1)
  * rake (11.3.0)
  * rspec (3.5.0)
  * rspec-core (3.5.4)
  * rspec-expectations (3.5.0)
  * rspec-mocks (3.5.0)
  * rspec-support (3.5.0)
  * rubocop (0.47.1)
  * ruby-progressbar (1.8.1)
  * simplecov (0.12.0)
  * simplecov-html (0.10.0)
  * slop (3.6.0)
  * term-ansicolor (1.4.0)
  * thor (0.19.4)
  * thread_safe (0.3.6)
  * tins (1.13.2)
  * tzinfo (1.2.2)
  * unicode-display_width (1.1.3)
  * xcodeproj (1.4.2)
```

gem/spec/spec_helper.rb
wuhao5: `require` look up files in `$LOAD_PATH`, rspec set up the local lib to be part of load path so when you do `require 'earlgrey'`, it will look up your local lib first.

`require_relative` usually requires you to maintain the same folder structure so once you move your code around, you need to make those changes. besides, logically, this is to test the gem, not your local `cli.rb` file. I'd argue `require_relative` is only used inside a module when you know exactly where your files are

wuhao5:
@bootstraponline do you want to double check if the fix works for you? if you are brave, you can run on the other of your projects that uses carthage and see how it goes? 🤓

gem/Gemfile
bootstraponline: run this command & let me know:

`bundle exec thor spec`

gem/lib/earlgrey/configure_earlgrey.rb
bootstraponline: that's why the comment is backwards compatible.

gem/spec/spec_helper.rb
bootstraponline: > logically, this is to test the gem, not your local cli.rb file.

no, the point is to test the on disk version of the code & not the installed gem version.

we only ever want to test the local version, why use require?

gem/Gemfile
wuhao5: I am getting an error:
```
Could not find command "spec".
```
I think your env is different than mine. what does your Gemfile.lock look like? you error seem to be you have missing gems, I assume you tried `bundle install` one more time?

gem/spec/spec_helper.rb
bootstraponline: > require_relative usually requires you to maintain the same folder structure so once you move your code around, you need to make those changes. 

nope... that is why we require by gem name. Only if the gem name changes will this statement be impacted.

> require_relative '../lib/earlgrey'

gem/Gemfile
bootstraponline: update: as the gem isn't actually using thor for task running / coveralls for code coverage then it's fine to keep these deleted.

bootstraponline:
> @bootstraponline do you want to double check if the fix works for you? if you are brave, you can run on the other of your projects that uses carthage and see how it goes?

I tried it on one of my projects and am getting a crash that repros only with this PR.

- `git clone https://github.com/bootstraponline/react_native_automation`
- `cd button/ios`
- `earlgrey install -t buttonTests`

```
$ earlgrey install -t buttonTests
Checking and Updating /code/other/react_native_automation/react_native_automation/button/ios/button.xcodeproj for EarlGrey.
Adding EarlGrey Framework Location as an Environment Variable
in the App Project's Test Target's Scheme Test Action button.
DYLD_INSERT_LIBRARIES is already set up for buttonTests, ignored.
/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/earlgrey-0.2.0/lib/earlgrey/configure_earlgrey.rb:300:in `block in add_carthage_search_paths': undefined method `|' for "$(SRCROOT)/Carthage/Build/iOS":String (NoMethodError)
	from /.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/earlgrey-0.2.0/lib/earlgrey/configure_earlgrey.rb:297:in `each'
	from /.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/earlgrey-0.2.0/lib/earlgrey/configure_earlgrey.rb:297:in `add_carthage_search_paths'
	from /.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/earlgrey-0.2.0/lib/earlgrey/configure_earlgrey.rb:154:in `configure_for_earlgrey'
	from /.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/earlgrey-0.2.0/lib/earlgrey/cli.rb:49:in `install'
	from /.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/thor-0.19.4/lib/thor/command.rb:27:in `run'
	from /.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/thor-0.19.4/lib/thor/invocation.rb:126:in `invoke_command'
	from /.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/thor-0.19.4/lib/thor.rb:369:in `dispatch'
	from /.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/thor-0.19.4/lib/thor/base.rb:444:in `start'
	from /.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/earlgrey-0.2.0/bin/earlgrey:20:in `<top (required)>'
	from /.rbenv/versions/2.3.3/bin/earlgrey:22:in `load'
	from /.rbenv/versions/2.3.3/bin/earlgrey:22:in `<main>'
```

gem/spec/spec_helper.rb
wuhao5: I see your point. we do test local version, as you would your test code work outside the lib just like how rspec set up test environment. I think the confusion is `rspec` is set up in a way that your lib behaves like a gem. I personally don't like to use `require_relative` as it needs more maintenance and lengthy unfriendly path.


also what if you tests code in a deeper folder, and you need a lot of `../`, and you decide to restructure your tests code?

bootstraponline: > also what if you tests code in a deeper folder, and you need a lot of ../, and you decide to restructure your tests code?

`spec/spec_helper.rb` doesn't change locations per standard rspec conventions, the rest of the test code can be restructured without impacting this path.

the only scenario where I see this breaking is if we rename the gem.

wuhao5: I guess what I am trying to say is those two are two separate components, and `require_relative` couples them. I am still not convinced :)

wuhao5:
reverted some of changes, @bootstraponline can you try it one more time?

gem/spec/spec_helper.rb
bootstraponline: I'm not convinced either. `require_relative` guarantees we're testing what we intend to test, not whatever happens to be on the load path.

the only coupling here is that the gem name is expected to remain constant. there's no other practical loss.

wuhao5: how about those:

[fastlane](https://github.com/fastlane/fastlane/search?l=Ruby&q=require_relative&type=&utf8=%E2%9C%93)
[CocoaPods](https://github.com/CocoaPods/CocoaPods/search?utf8=%E2%9C%93&q=require_relative&type=)
[Carthage](https://github.com/Carthage/Carthage/search?utf8=%E2%9C%93&q=require_relative&type=)
[Rails](https://github.com/rails/rails/search?l=Ruby&p=4&q=require_relative&type=&utf8=%E2%9C%93)
[Rake](https://github.com/ruby/rake/search?utf8=%E2%9C%93&q=require_relative&type=)
[Bundler](https://github.com/bundler/bundler/search?utf8=%E2%9C%93&q=require_relative&type=)
This list can go on and on...

None of those are either using `relative_require` or require it outside the test scope. I wouldn't want earlgrey to be the first project that does this practice. :/

Yes, it does massage your load_path, but that's based on top of knowledge of using RSpec, and you would expect RSpec would test your local lib/gem, not something else; if it does, it's a bug for RSpec.

Would you run those spec.rb directly? probably not.

bootstraponline: as you pointed out, require relative is commonly used in the test context. I don't think we'd be the first.

Earlgrey isn't defined in the lib folder & doesn't import the related files then that's a bug the tests should catch. I wouldn't want the require call to succeed if that file has been moved or the contents were deleted.

wuhao5: Discussed offline and understood pros and cons, but we are still not convinced by each other, but as this is a small issue and I'd like to follow what other ruby projects are doing, we'll just keep it as is now and we are both fine with it.

wuhao5:
I have rebased the commits so it's cleaner, it should be ready to be merged.

khandpur:
Would be good to get a few more of our users to pilot this before releasing. Also, please remember to send an announcement on earlgrey-discuss + slack about this change, how it affects the current users who are using the older version of the gem and the migration process for them.

gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -13,19 +14,50 @@
...
-# global export for unqualified use in Pods file
-def configure_for_earlgrey(*args)
-  EarlGrey.configure_for_earlgrey(*args)
+require 'colored'
+require 'fileutils'
+require 'xcodeproj'
+
+def configure_for_earlgrey
+  raise <<-S
+    configure_for_earlgrey in post_install is now deprecated, you should safely please it now.
```
bootstraponline: `you should safely please it now.`
?

maybe instead of immediately breaking everyone by raising an exception, we should only print a warning and then call through to `EarlGrey.configure_for_earlgrey`? Then in the next release we can drop the global method entirely.

wolffan:
I wasn't able to install EarlGrey from a project nor from the Demo project.
In this branch I can now! 

Many thanks!

gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -34,64 +66,59 @@ def path_for(project_name, ext)
...
 
     # Raise error message after removing excessive spaces.
-    # @param message [String] the message to raise
+    # @param [String] message the message to raise
     # @return [nil]
     def error(message)
       raise strip(message)
     end
 
     # Prints string as blue after stripping excess spacing
```
bootstraponline: `Prints string as blue` 😭 

looks like the comment needs to be updated

bootstraponline:
I tried the gem on a few of our projects and everything appears to be working.

gem/lib/earlgrey/configure_earlgrey.rb
wuhao5: sounds good!

Demo/EarlGreyExample/Podfile
```diff
@@ -18,38 +18,19 @@ platform :ios, '8.0'
...
-
-  # For each test target, you need to call the EarlGrey script's edit method once.
-  TARGETS.each do |test_target|
-    var_args = {}
-    if test_target.include? "Swift"
-      var_args = {swift: true}
-    end
-    configure_for_earlgrey(installer, PROJECT_NAME, test_target, test_target + '.xcscheme', var_args)
-  end
+  inherit! :search_paths
```
tirodkar: Shouldn't this be on top of pod EarlGrey ?

wuhao5:
I reverted the doc for this PR and fixed the Podfile diff. I think this PR is good to go, and I can push the gem after the PR is in later today if things move smoothly.

wuhao5:
> I thought we don't need this with the new gem changes?

right, but not before we push the gem. I wanted to make a separate PR for this once the gem is pushed and confirmed to work besides @tirodkar is also working on the updating the doc. sounds good?

bootstraponline:
> right, but not before we push the gem. I wanted to make a separate PR for this once the gem is pushed and confirmed to work besides @tirodkar is also working on the updating the doc. sounds good?

yes

khandpur:
Some issues here with runs on Travis:
https://travis-ci.org/google/EarlGrey/jobs/216563453
https://travis-ci.org/google/EarlGrey/jobs/216563454

Before we can merge this in, let's make sure our tests are happy. Could you please address the above and do a rerun of just the relevant targets?

wuhao5:
I cancelled a few runs earlier where I missed the failure when running `pod install` from non-project folder, I didn't look into why it was working before this PR, but the fix is to copy the file into the group folder other than depending on the `installer.config.installation_root`. also I reverted the podfile as the gem should be backward compatible. there was also an issue in the old script where it tried to install the "latest" gem (which would be 0.1.6), but the 0.2.0 is already being loaded (from pod command line). anyhow, we removed all those uncommon codes.

the new travis now is all green

PR #473, opened 2017-03-30
Update the doc and Podfile for the gem 0.2.0

wuhao5:


wuhao5:
I cancelled the travis since I haven't pushed the gem yet. will restart it later

tirodkar:
For updating the gem version, you might want to bump the gem number as well in this PR.

Demo/EarlGreyExample/Podfile
```diff
@@ -37,19 +34,3 @@ target "EarlGreyExampleSwiftTests" do
   pod 'EarlGrey'
```
tirodkar: Add a comment stating that we don't need to add any other flag or so for Swift etc.

PR #475, opened 2017-03-31
Add an API to dismiss the keyboard from EarlGrey & Provide some documentation fixes

tirodkar:


EarlGrey/Core/GREYKeyboard.m
```diff
@@ -36,9 +37,9 @@
 static GREYTapAction *gTapKeyAction;
 
 /**
- *  Flag set to @c YES when the keyboard is shown, @c NO when keyboard is hidden.
+ *  Flag set to @c true when the keyboard is shown, @c false when keyboard is hidden.
  */
-static BOOL gIsKeyboardShown = NO;
+static atomic_bool gIsKeyboardShown = false;
```
sid-github: Since this just a value var, does using volatile work? You can declare this as volatile and then use the variable normally:
static volatile BOOL gIsKeyboardShown = NO;

tirodkar: TTBOMK, when I looked at this, `volatile` will only ensure that the compiler knows that other threads can modify the variable, thus stopping compiler optimizations, whereas `atomic` ensures that each operation on it is atomic in nature, thus providing greater thread-safety. Do you think there's another reason for using `volatile`?

sid-github: I actually take back my comment: Sundeep just shared [this](http://stackoverflow.com/questions/2484980/why-is-volatile-not-considered-useful-in-multithreaded-c-or-c-programming) which taks about this in more detail, also from [this](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/Multithreading/ThreadSafety/ThreadSafety.html) its clear that we need `OSMemoryBarrier` too inaddition to `volatile`.

tirodkar: [Here's](http://stackoverflow.com/questions/31535840/which-scenes-keyword-volatile-is-needed-to-declare-in-objective-c) also a good read on `volatile`.

wuhao5: this is cool, good to know, just be mindful that `stdatomic.h` is C11 standard (just looked, xcode7 should have already supported it), but I like the fact we are chasing the latest standard.

khandpur:
Fixes 472

Issue #479, opened 2017-04-04
matcherForEnabledElement is treating nil elements as enabled


RichardGuion:
When using grey_enabled in a test I noticed that any text I used did not assert

```
        EarlGrey.select(elementWithMatcher:
            allOf(withText("bar"),
                  withAncestor(allOf(withClass(MyClass.self), isVisible())))).assert(grey_enabled())
```

This is a case where "bar" did not exist at all and I was expecting an assertion thrown.

The answer is here: 
https://github.com/google/EarlGrey/blob/b0ed5e03fb2b5a5ba00843811c9e87baa547bf6b/EarlGrey/Matcher/GREYMatchers.m#L476

The default value of BOOL matched should be NO instead of YES.

khandpur:
Note that the returned matcher should also contain a `isKindOfClass` matcher for UIControl class types:
```
return grey_allOf(grey_kindOfClass([UIControl class]), isEnabledMatcher, grey_not(grey_ancestor(grey_not(isEnabledMatcher))), nil);
```

khandpur:
@RichardGuion would you like to take a stab at it?

PR #481, opened 2017-04-05
matcherForEnabledElement default matched value should be NO/false (#479)

RichardGuion:


RichardGuion:
I signed it! The CLA.

tirodkar:
@RichardGuion could you confirm if you signed the CLA with the same email as when committing?

PR #484, opened 2017-04-12
Fixing a bug in interaction point calculation.

khandpur:
Fix bug in visibility checker around interaction point. It is possible that neither the activation point nor the center of an element's visible area is actually visible even though, some part of the element is visible and interact-able by the user. To fix that, we use a default point guaranteed to be visible. The default point is selected from a set of visible point returned by performing pixel shift check.

### DIFF
--- a/EarlGrey/Additions/UIViewController+GREYAdditions.m
+++ b/EarlGrey/Additions/UIViewController+GREYAdditions.m
@@ -27,7 +27,7 @@
  *  The class for UICompatibilityInputViewController which isn't tracked here since we've faced
  *  issues with tracking it when it comes to typing on keyboards with accessory views.
  */
-static Class compatibilityVCClass;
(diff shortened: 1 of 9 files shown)


EarlGrey/Common/GREYAnalyticsDelegate.h
```diff
@@ -25,11 +25,11 @@
  *
  *  @note In case of failure to track the method must fail silently to prevent test interruption.
  *
- *  @param trackingID The tracking ID under which to track this event.
- *  @param clientID   The ID for the user sending this event.
- *  @param category   The Event Category for the event hit.
- *  @param action     The Event Action for the event hit.
- *  @param value      The event value for the event hit.
+ *  @param trackingID  The tracking ID under which to track this event.
```
sid-github: Please revert this.

PR #489, opened 2017-04-13
Fix Swift breaks by correcting issue with Gem adding EarlGrey.swift irrespective of the type of the project.

tirodkar:
Also deletes unneeded Swift App imports in the project which was causing the ObjC tests to also contain EarlGrey.swift.

### DIFF
--- a/Demo/EarlGreyExample/EarlGreyExample.xcodeproj/project.pbxproj
+++ b/Demo/EarlGreyExample/EarlGreyExample.xcodeproj/project.pbxproj
@@ -11,7 +11,6 @@
 		2CB7315A1C29E8E600CF35C1 /* LaunchScreen.xib in Resources */ = {isa = PBXBuildFile; fileRef = 5F9961071AE0CF4F0034F503 /* LaunchScreen.xib */; };
 		2CB7315B1C29E8F400CF35C1 /* Main.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 5F5A53591ADE670C00F81DF0 /* Main.storyboard */; };
 		2CE7FA481C35AAB00079DE2F /* ViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2CE7FA471C35AAB00079DE2F /* ViewController.swift */; };
-		2CE7FA491C35AAB00079DE2F /* ViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2CE7FA471C35AAB00079DE2F /* ViewController.swift */; };
 		5F5A537E1ADE67D500F81DF0 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F5A537D1ADE67D500F81DF0 /* AppDelegate.swift */; };
 		5F99610B1AE0CF4F0034F503 /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 5F9961061AE0CF4F0034F503 /* Images.xcassets */; };
 		5FDE055D1B0DAA090037B82F /* EarlGreyExampleTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 5FDE055C1B0DAA090037B82F /* EarlGreyExampleTests.m */; };
@@ -296,7 +295,6 @@
 			isa = PBXSourcesBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
-				2CE7FA491C35AAB00079DE2F /* ViewController.swift in Sources */,
 				5FDE055D1B0DAA090037B82F /* EarlGreyExampleTests.m in Sources */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
@@ -532,6 +530,7 @@
 				SWIFT_OBJC_BRIDGING_HEADER = "";
 				SWIFT_OBJC_INTERFACE_HEADER_NAME = "EarlGreyExampleSwift-Swift.h";
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_VERSION = "";
 				TEST_HOST = "$(BUILT_PRODUCTS_DIR)/EarlGreyExampleSwift.app/EarlGreyExampleSwift";
 			};
 			name = Debug;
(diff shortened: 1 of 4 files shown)


gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -353,7 +354,21 @@ def add_earlgrey_framework(target, framework_ref)
...
       target.source_build_phase.files_references.any? do |ref|
-        SWIFT_FILETYPE == (ref.last_known_file_type || ref.explicit_file_type)
+        file_type = ref.last_known_file_type
+        if file_type.to_s.empty?
+          file_type = ref.explicit_file_type
+        end
+
+        if SWIFT_FILETYPE == file_type
+          return true
+        elsif OBJC_FILETYPE == file_type
```
wuhao5: why would you compare objc? it can have both objc and swift source files, not?

Demo/EarlGreyExample/EarlGreyExample.xcodeproj/project.pbxproj
```diff
@@ -532,6 +530,7 @@
 				SWIFT_OBJC_BRIDGING_HEADER = "";
 				SWIFT_OBJC_INTERFACE_HEADER_NAME = "EarlGreyExampleSwift-Swift.h";
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_VERSION = "";
```
wuhao5: revert this?

Demo/EarlGreyExample/EarlGreyExample.xcodeproj/project.pbxproj
```diff
@@ -562,6 +561,7 @@
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SWIFT_OBJC_BRIDGING_HEADER = "";
 				SWIFT_OBJC_INTERFACE_HEADER_NAME = "EarlGreyExampleSwift-Swift.h";
+				SWIFT_VERSION = "";
```
wuhao5: revert this as well

wuhao5: i think you should remove the default value from the project level. unspecified can also indicate it has swift sources, right?

gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -364,8 +379,7 @@ def has_swift?(target)
     # @param [Xcodeproj::Project] project
     # @param [PBXNativeTarget] target
     def copy_swift_files(project, target, swift_version = nil)
-      return unless has_swift?(target) || !swift_version.nil?
-
+      return unless has_swift?(target) || !swift_version.to_s.empty?
```
wuhao5: I don't know the semantic meaning of Unspecified or empty.

my guess is if you have swift files but doesn't specify its version, it would throw an error when compiling?

Scripts/travis.sh
```diff
@@ -108,8 +108,9 @@ elif [ "${TYPE}" == "EXAMPLE_PODS_SWIFT" ]; then
   cd gem/
   bundle install
   rake install
-  cd ..
-  pod install --project-directory=Demo/EarlGreyExample
```
wuhao5: does it not work any more?

gem/lib/earlgrey/configure_earlgrey.rb
wuhao5: I think this actually doesn't make sense if you look at the method name. you can have a separate method like has_objc?, but I still don't know if you really need to check that.

also throwing the error doesn't seem to be a good idea here, has_swift? returns false if it doesnt have swift files.

tirodkar: The method still checks if Swift is present. Just ensures that something is explicitly set.

You can have ObjC and Swift in the same target as we did with EGExampletests. But it came up as file type - Swift. I don't think both can be nil.

Demo/EarlGreyExample/EarlGreyExample.xcodeproj/project.pbxproj
tirodkar: On changing the value at the project level, I received the following error:

```
“Swift Language Version” (SWIFT_VERSION) is required to be configured correctly for targets which use Swift. Use the [Edit > Convert > To Current Swift Syntax…] menu to choose a Swift version or use the Build Settings editor to configure the build setting directly.
```

Scripts/travis.sh
tirodkar: Moved this to a separate CL. Reverted.

gem/lib/earlgrey/configure_earlgrey.rb
tirodkar: That's correct. Hmmm... should change this to an & then?

gem/lib/earlgrey/configure_earlgrey.rb
wuhao5: The logic I am reading here:
go thru each file type, if one objc encounters, return false; if one swift file encounters, return true.

so, let's say you have: [a.m, b.swift], it will return false, where it should return true.

I don't quite follow the reason behind...

Demo/EarlGreyExample/EarlGreyExample.xcodeproj/project.pbxproj
wuhao5: I see, I thought this target doesn't use swift.

gem/lib/earlgrey/configure_earlgrey.rb
tirodkar: I just re-tested this out. This method does indeed seem to be working correctly.

The issue was the version number, even though it was set to zero, was being returned as true, which led to the Swift bundling being done regardless. I'll change this back!

tirodkar: Done.

Demo/EarlGreyExample/EarlGreyExample.xcodeproj/project.pbxproj
tirodkar: It used to have a Swift file, which is why it was set to 3.0. Without this, it'll inherit from the parent project which has 3.0, which will trigger the Swift File Bundling in configure_earlgrey..., which will cause the following error:

```
RuntimeError - Bundled swift missing
/Users/tirodkar/.rvm/gems/ruby-2.3.1/gems/earlgrey-0.2.0/lib/earlgrey/configure_earlgrey.rb:382:in `copy_swift_files'
```

gem/lib/earlgrey/configure_earlgrey.rb
```diff
@@ -41,6 +41,7 @@ module EarlGrey
   HEADER_SEARCH_PATHS = 'HEADER_SEARCH_PATHS'.freeze
 
   SWIFT_FILETYPE = 'sourcecode.swift'.freeze
+  OBJC_FILETYPE = 'sourcecode.c.objc'.freeze
```
wuhao5: revert this.

tirodkar: Done

PR #474, opened 2017-03-31
Better Swift support for grey_allOf and grey_anyOf. Removing the Swift 2 wrapper methods to ease migration to Swift3.

tirodkar:

PR #490, opened 2017-04-13
Change Pod Install to Pod Update. Without this, we're probably testing on the same Pods version.

tirodkar:
…

### DIFF
--- a/Scripts/travis.sh
+++ b/Scripts/travis.sh
@@ -108,7 +108,9 @@ elif [ "${TYPE}" == "EXAMPLE_PODS_SWIFT" ]; then
   cd gem/
   bundle install
   rake install
-  cd ..
+  cd ../Demo/EarlGreyExample
+  pod update
+  cd ../../
   pod install --project-directory=Demo/EarlGreyExample
   execute_xcodebuild Demo/EarlGreyExample/EarlGreyExample.xcworkspace EarlGreyExampleSwiftTests
 elif [ "${TYPE}" == "CARTHAGE" ]; then


Scripts/travis.sh
```diff
@@ -108,7 +108,9 @@ elif [ "${TYPE}" == "EXAMPLE_PODS_SWIFT" ]; then
   cd gem/
   bundle install
   rake install
-  cd ..
+  cd ../Demo/EarlGreyExample
+  pod update
```
wuhao5: if this is on travis, are you sure Travis won't clean up after it's done?

also you can consider `pod update --project-directory=` after `pod install`

tirodkar: If we don't add any new pods, pod update would work without a Pod install. Travis doesn't seem to clean when its done, which is why we kept getting the swift bug.

wuhao5: `pod update` should always happen after `pod install`, right?

I am curious to see the travis log, it sounds alarming to me that every run in Travis is different, at lease in a way they cache stuff from others...

tirodkar: Not necessarily.
Pod install is only needed when you've added a new pod.
https://guides.cocoapods.org/using/pod-install-vs-update.html

Here's a failing job: 
https://travis-ci.org/google/EarlGrey/jobs/221570274

wuhao5: the error seems to be a legit error after #474 , grey_allOf accepting varargs but grey_allOfMatchers accepting an array, is it not the case here?

tirodkar: The API was updated.  If it was an error with EG or the tests, then FTRSwift tests would have failed as well. I'd surmise it's failing here but not failing after a Pod Update is because it is using a stale pods version. We aren't guaranteed a clean build by travis if I'm right.

PR #504, opened 2017-05-05
Fixing styling issues in install scripts

khandpur:


khandpur:
I am okay with these commits being part of this project.

Issue #510, opened 2017-05-23
Adding EarlGrey Pod to an existing project with R.swift breaks the build


mariozel:
Currently I'm working on a project where we use cocoapods and one of which is R.swift. Adding EarlGrey to this project breaks the build with the following error:
`/MyNewApp/build/MyNewApp.build/Debug-iphonesimulator/MyNewApp.build/Script-5DB000921ED488FE0011D379.sh: line 5: 59938 Illegal instruction: 4  "$PODS_ROOT/R.swift/rswift" "$SRCROOT"
Command /bin/sh failed with exit code 132`

I've created a sample (empty) project in order to reproduce this issue successfully. https://github.com/mariozel/MyNewApp

To Repro, clone the project, run pod install then try to build.

If anyone has any ideas what could be happening please let me know.

khandpur:
Thanks for adding a repro project. It really helps debugging install issues.

@wuhao5 Can you take a look at this one?

wuhao5:
I am not able to compile your project even after removing the EarlGrey pod. I haven't used R.swift before, but even I tried to run `rswift`, it errors too.

wuhao5:
I am able to repro the issue: looks like R.swift is looking at Products group and we are adding EarlGrey.framework there. The workaround is not trivial: you will need to remove it from the xcodeproj file manually (Xcode will not allow you do so) and add it to another group (i.e. frameworks) and fix the EarlGrey copy file script where it links to this framework.

I will work on the fix.

khandpur:
@wuhao5 thanks for looking into it. Any ETA? Will we be able to release it on Friday?

wuhao5:
It will be a gem update, possible independent of a separate release from the EarlGrey mainstream. it should be done in a day.

mariozel:
NIce! Thanks lads.

wuhao5:
the new workaround: patch the PR and do `rake install` under `gem/`, and then do pod install one more time.

PR #511, opened 2017-05-23
Fix source_tree and relative path

wuhao5:
https://github.com/google/EarlGrey/issues/510

### DIFF
--- a/gem/lib/earlgrey/configure_earlgrey.rb
+++ b/gem/lib/earlgrey/configure_earlgrey.rb
@@ -48,9 +48,9 @@ module EarlGrey
 
   EARLGREY_FRAMEWORK = 'EarlGrey.framework'.freeze
   CARTHAGE_FRAMEWORK_PATH =
-    '${SRCROOT}/Carthage/Build/iOS/EarlGrey.framework'.freeze
+    'Carthage/Build/iOS/EarlGrey.framework'.freeze
   POD_FRAMEWORK_PATH =
-    '${SRCROOT}/Pods/EarlGrey/EarlGrey/EarlGrey.framework'.freeze
+    'Pods/EarlGrey/EarlGrey/EarlGrey.framework'.freeze
 
   class << self
     attr_reader :project_name, :installer, :test_target, :test_target_name,
@@ -264,13 +264,13 @@ module EarlGrey
                          POD_FRAMEWORK_PATH
                        end
 
-      framework_ref = project.products_group.files.find do |f|
+      framework_ref = project.frameworks_group.files.find do |f|
         # TODO: should have some md5 check on the actual binary
         f.path == framework_path
       end
       unless framework_ref
-        framework_ref = project.products_group.new_file(framework_path)
-        framework_ref.source_tree = 'SRCROOT'
+        framework_ref = project.frameworks_group.new_file(framework_path)
+        framework_ref.source_tree = 'SOURCE_ROOT'
       end
       framework_ref
     end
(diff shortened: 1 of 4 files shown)


tirodkar:
@bootstraponline, do test this out if you can for Carthage!

gem/lib/earlgrey/version.rb
```diff
@@ -14,5 +14,5 @@
 #  limitations under the License.
 
 module EarlGrey
-  VERSION = '0.2.1'.freeze unless defined? ::EarlGrey::VERSION
+  VERSION = '0.2.2'.freeze unless defined? ::EarlGrey::VERSION
```
bootstraponline: 😭

bootstraponline:
> @bootstraponline, do test this out if you can for Carthage!

I'll test it out now.

gem/lib/earlgrey/version.rb
wuhao5: Hah, sorry I forgot what was the decision to be made, is it to be sync'd or not?

Note that we are doing more mainstream releases than gem releases.

bootstraponline: can we just make it `1.9.2`? `0.2.2` is not conveying any useful info to the end user

bootstraponline:
I verified this works on the latest version of carthage.

```
gem build earlgrey.gemspec
gem list earlgrey
gem install --local earlgrey-0.2.2.gem
gem list earlgrey
cd Demo/EarlGreyExample
brew update
brew upgrade carthage
earlgrey install --carthage --swift-version=3.0 --target=EarlGreyExampleSwiftTests
carthage update EarlGrey --platform ios
open EarlGreyExample.xcodeproj
```

khandpur:
lots of diff detected here: https://travis-ci.org/google/EarlGrey/jobs/235431968

khandpur:
@bootstraponline thanks for testing!

khandpur:
@wuhao could we perhaps add a test for this with R.swift? That way we can catch regressions in the future?

wuhao5:
@khandpur looks like you mentioned the wrong person :)

forgot to run rake test, good catch, will also add those tests too. the only concern is if we should support R.swift and add another dependency here. I'd say R.swift is doing something incompatible with the unusual paths which have some special characters. R.swift should ignore or warn, rather than just fail.

gem/lib/earlgrey/version.rb
wuhao5: I am dropping this now and update it with the release schedule.

wuhao5:
Fixed the fixtures and added R.swift to the example project. waiting for CI to finish.

Demo/EarlGreyExample/Podfile
```diff
@@ -30,6 +30,9 @@ end
 target "EarlGreyExampleSwiftTests" do
   project 'EarlGreyExample'
 
+  # Make sure it is compatible with other pods, i.e. R.swift
+  pod 'R.swift'
```
tirodkar: Do we have any R.swift code in this target? We use EGExample / EGExampleSwift for Sanity Testing our new pods. Might not want to add a new dependency on it.

tirodkar: I'd create a new target or add it to the Carthage tests or so? Maybe run some R.swift code to it?

wuhao5: reverted this for now.

tirodkar:
Waiting on CI. Thanks.

PR #519, opened 2017-05-30
Added Auto Layout support to some nibs. Adjusted constraints for some nibs.

saurabhj80:


saurabhj80:
I signed it!

khandpur:
Can you please update the .xibs so they're compatible with older versions of Xcode. There's an option on the xib interface to do that.

tirodkar:
Will wait on the CI before merging.

PR #529, opened 2017-06-06
Use Chrome 59 for PNG cheatsheet rendering

bootstraponline:
- Zero dependencies to render a PNG now.
- PDF rendering should work also however Chrome 59 doesn't capture the text. Seems like a bug.


Scripts/cheatsheet/render.rb
```diff
@@ -14,34 +14,22 @@
...
 
-  # remove right hand scrollbars by cropping.
-  crop_width = screenshot.width - 20
-  screenshot.crop!(0, 0, crop_width, screenshot.height)
-  screenshot.save(cheatsheet_png, :best_compression)
-ensure
-  web_driver.quit
-end
+# run_command "#{chrome} --headless --disable-gpu --print-to-pdf=#{cheatsheet_pdf} file://#{cheatsheet_html}"
+run_command "#{chrome} --headless --hide-scrollbars --disable-gpu --screenshot=#{cheatsheet_png} --window-size=1024,2550 file://#{cheatsheet_html}"
```
wuhao5: does this require version 59? can you also add a check and assert that too?

bootstraponline: > does this require version 59?

yes

> can you also add a check and assert that too?

sure

Scripts/cheatsheet/render.rb
```diff
@@ -14,34 +14,21 @@
...
-  screenshot = eyes_browser.fullpage_screenshot
-
-  # remove right hand scrollbars by cropping.
-  crop_width = screenshot.width - 20
-  screenshot.crop!(0, 0, crop_width, screenshot.height)
-  screenshot.save(cheatsheet_png, :best_compression)
-ensure
-  web_driver.quit
-end
+# run_command "#{chrome} --headless --disable-gpu --print-to-pdf=#{cheatsheet_pdf} file://#{cheatsheet_html}"
```
wuhao5: remove this line?

Scripts/cheatsheet/render.rb
wuhao5: cool! I patched and tested it locally, worked for me!

bootstraponline:
> will it make more sense to rewrite this in bash?

if we use bash then i'd prefer a one liner and dropping all the conditional checking. fine with that?

wuhao5:
>  a one liner and dropping all the conditional checking
Yes, I like one liner! I think you could still do some checks just so when anyone runs it, it can give meaningful errors.

also @sid-github or @tirodkar , WDYT?

bootstraponline:
Ok, added bash version.

PR #530, opened 2017-06-07
Add RVM install step in before install for travis runs.

tirodkar:
…≈


khandpur:
Could we also open an issue to track this?

tirodkar:
https://github.com/google/EarlGrey/issues/531

PR #547, opened 2017-06-29
Improved visibility checker to fix https://github.com/google/EarlGrey…

emitch:
…/issues/532 and added corresponding tests. Also resolved an issue where the hierarchy dump functional test would fail incorrectly on old iOS versions.


emitch:
I signed it!

PR #549, opened 2017-06-29
Resolved visibility checker overlapping view issue.

### LINKED ISSUE #532: Actions are applied to overlapping views
If there's a scrollView `A` overlapped with a scrollview `B` at the bottom, then a scroll-down action applied to the `A` element generates touches at the bottom of it and thus `B` is scrolled instead.

If there's a button `A` overlapped with a button `B` at the center, then a tap action on `A` causes an exception.

An example project for these cases with the tests: [https://github.com/wix/earlgrey-with-overlapping-interactable-views](https://github.com/wix/earlgrey-with-overlapping-interactable-views)

I think that there should be a check of the area available for tap and scroll (the first touch point). Something like this (for tap): going over all the views that appear above the view and subtracting areas covered by other views that would catch the touches/gestures instead of the view. Then the action point should be a point in the area. If there's too small area left, then it may be good to throw an exception (it would be nice if the minimum amount/percent of space is configurable) - because it's not enough area for a real user to tap.

Another (simpler to implement) option is to have a list of possible locations for touch points and checking only those points (checking if the view that is going to accept scroll at the bottom center of the view is indeed the view itself and if not, try switching to alternative locations).

emitch:
Fixed #532 and a minor issue in the window hierarchy functional test causing incorrect failures on older iOS versions.

### DIFF
--- a/EarlGrey/Common/GREYVisibilityChecker.m
+++ b/EarlGrey/Common/GREYVisibilityChecker.m
@@ -690,6 +690,73 @@ inline void GREYVisibilityDiffBufferSetVisibility(GREYVisibilityDiffBuffer buffe
 }
 
 /**
+ *  Given a list of values representing a histogram (values are the heights of the bars), this
+ *  method returns the largest contiguous rectangle in that histogram.
+ *
+ *  @param histogram The array of values representing the histogram.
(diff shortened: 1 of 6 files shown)


emitch:
I signed it!

PR #575, opened 2017-07-14
Sync 20170714

wuhao5:


wuhao5:
missed one added file, rebased and running the test again.

wuhao5:
only saw the flaky web tests failures, it should be good to go.

PR #578, opened 2017-07-19
Fix typo in the sample code to match the API

gimite:
The method is defined here:
https://github.com/google/EarlGrey/blob/master/EarlGrey/Action/GREYActionBlock.h#L61

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -268,7 +268,7 @@ of a block. The following code creates an action using a block that invokes a cu
 ```objc
 - (id<GREYAction>)animateWindowAction {
   return [GREYActionBlock actionWithName:@"Animate Window"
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -268,7 +268,7 @@ of a block. The following code creates an action using a block that invokes a cu
 ```objc
 - (id<GREYAction>)animateWindowAction {
   return [GREYActionBlock actionWithName:@"Animate Window"
-                              constraint:nil
+                              constraints:nil
```
tirodkar: Formatting was off here. That's cool. Thanks for the fix @gimite!

gimite: Oops sorry. Sent a PR to fix it: #579

Issue #580, opened 2017-07-20
Missing Swift 3.2 and Swift 4.0 Support 


mettieDie:
I updated my Project to Swift 4.0. Know CocoaPods crashed because EarlGrey doesn't support Swift 4.0

cipherCOM:
Same on my side. `pod install` doesn't work anymore due to missing swift bundles.

tirodkar:
I haven't looked into adding Swift 4.0 support as of now. Can you possibly point to what we would need to do for this?

wuhao5:
did you see the error message like this: `Bundled swift missing`?

mettieDie:
Yes exactly this error Message. The Folders for Swift 3.2 and Swift 4.0 are Missing.I Have made a Pull Request with a Fix.

khandpur:
@mettieDie Do you plan on getting the comments in the PR resolved anytime this week? We could send another PR to solve this bug as it is preventing people from running `pod install`

MaxJZimmermann:
@khandpur  +1 Would be great!

mettieDie:
I think it's better to send another PR Request. I have my last exam end of the week and no time to do it. sorry.

khandpur:
Sure, no worries @mettieDie. We will take care of this.

All the best for your last exam!

khandpur:
Should be fixed in #591

PR #581, opened 2017-07-20
Fix for Missing Swift 3.2 and Swift 4.0 Support Issue #580

mettieDie:
Created Swift 3.2 and Swift 4.0 Folders to ensure building Earlgrey for Projects using Swift 3.2 and Swift 4.0

### DIFF
--- a/gem/lib/earlgrey/files/Swift-3.2/EarlGrey.swift
+++ b/gem/lib/earlgrey/files/Swift-3.2/EarlGrey.swift
@@ -1,0 +1,158 @@
+//
+// Copyright 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


mettieDie:
I signed it!

mettieDie:
I signed it!

wuhao5:
If they are the same, can we make a change here: 
https://github.com/google/EarlGrey/blob/master/gem/lib/earlgrey/configure_earlgrey.rb#L381

so when the swift file is not found, use the backup 3.0 version?

mettieDie:
Yes they are the same. We can use the Swift 3.0 Version. I can implement this tomorrow.

tirodkar:
@bootstraponline Is it time for us to deprecate he 3.0 and 2.3 files for these? At least the 2.3 one?

bootstraponline:
> @bootstraponline Is it time for us to deprecate he 3.0 and 2.3 files for these? At least the 2.3 one?

I think so. We should probably only support the latest stable Xcode release. Swift is changing too rapidly to support all the versions forever.

khandpur:
@tirodkar @wuhao5 Could we merge this if it's good to go?

wuhao5:
I think the decision is to not add another swift file but have gem to handle it.

tirodkar:
Hello, I thought we were waiting on the additions to the gem here. Is that to be done in a separate PR?

wuhao5:
closing in favor of #591

PR #583, opened 2017-07-21
Modify the CAAnimation Delegate usage to prevent failing assert in iOS 11

tirodkar:
…


willlarche:
Just tested. It works! Thank you.

PR #618, opened 2017-08-22
Bumping version to 1.12.0

sskhandp:


EarlGrey.podspec.json
```diff
@@ -18,7 +18,7 @@
 // frameworks in the app bundle need to include bitcode.
 {
   "name": "EarlGrey",
-  "version": "1.11.0",
+  "version": "1.12.0”,
```
tirodkar: Should be:

"version": "1.12.0",

PR #630, opened 2017-09-12
Fix spacing to not break lint rules

kdawgwilk:
This keeps bothering me in our project where we use SwiftLint and would love to not have this file generated every time we fix the lint rules


kdawgwilk:
I signed it!

khandpur:
Thanks for sending us a fix!

PR #631, opened 2017-09-20
ScrollView without content, scroll to content edge issue simulation.

Ad1991:
Changes to simulate the UIScrollView issue where scroll view gets stuck in infinite loop if you try to scroll to content edge and scrollView's content size is equal to or smaller than scrollView's own size. For more check this [conversation](https://groups.google.com/forum/#!topic/earlgrey-discuss/JaM4g9b5QFg).

### DIFF
--- a/Tests/FunctionalTests/Sources/FTRScrollViewTest.m
+++ b/Tests/FunctionalTests/Sources/FTRScrollViewTest.m
@@ -28,6 +28,10 @@
   [self openTestViewNamed:@"Scroll Views"];
 }
 
+- (void)DISABLED_testFreezingScrollView {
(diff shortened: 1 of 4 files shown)


Tests/FunctionalTests/Sources/FTRScrollViewTest.m
```diff
@@ -28,6 +28,10 @@ - (void)setUp {
   [self openTestViewNamed:@"Scroll Views"];
 }
 
+- (void)DISABLED_testFreezingScrollView {
```
tirodkar: Why is this test disabled?

tirodkar: I see. This is replicating an issue for us. Can you link the issue as well / create one if not already there?

khandpur: I don't think we should block this PR on that?

Feel free to create and link one to this Adarsh.

Issue #639, opened 2017-10-07
Error in assertion inout parameter is not as full as in logs of failure handler


artyom-razinov:
There is a method:

`- (instancetype)assert:(id<GREYAssertion>)assertion error:(__strong NSError **)errorOrNil`

When I pass `nil` to `errorOrNil` default failure handler prints full error to log:

```
  "Action Name" : "Type 'ewewrwe@mail.'",
  "Element Description" : "<AvitoDesignKit.TextField:0x7ff2c189aa00; AX=N; AX.value='Электронная почта'; AX.frame={{16, 96}, {343, 60}}; AX.activationPoint={187.5, 126}; AX.traits='UIAccessibilityTraitNone'; AX.focused='N'; frame={{16, 0}, {343, 60}}; alpha=1; text=''>",
  "Failed Constraint(s)" : "!(isSystemAlertViewShown)",
  "All Constraint(s)" : "!(isSystemAlertViewShown)",
  "Recovery Suggestion" : "Adjust element properties so that it matches the failed constraint(s)."
```

Notice that there are "Failed Constraint(s)", it is very useful for debugging.

I want to have custom reporting system, so I pass pointer to `errorOrNil`, but when I print error it is not full, it has only action name and element description.

There is a code that removes fields that I need from actionError:

```
      NSMutableDictionary *errorDetails = [[NSMutableDictionary alloc] init];
      errorDetails[kErrorDetailActionNameKey] = action.name;
      errorDetails[kErrorDetailElementMatcherKey] = _elementMatcher.description;
      [(GREYError *)actionError setErrorInfo:errorDetails];
```

I propose to change it to:

```
      GREYError *greyError = (GREYError *)actionError;
      NSMutableDictionary *errorDetails = [[NSMutableDictionary alloc] initWithDictionary:greyError.errorInfo];
      errorDetails[kErrorDetailElementMatcherKey] = _elementMatcher.description;
      [greyError setErrorInfo:errorDetails];
```

Everything will be same except all fields will be available in `errorOrNil`.

Issue #31, opened 2016-02-25
Some keys can not be tapped on non English key Keyboard.


lestathc:
Hi,

I tried the EarlGrey, if you want to test non-English app, some keys can not be tapped, like "space". I tried with "fr", the space key should be "espace".

Besides of that, EarlGrey is great, thanks for making this.

tirodkar:
@lestathc you should be able to type any language string using grey_replaceText() now. Please do give it a try.

KazuCocoa:
In my case, I could put some Japanese into search field with `grey_replaceText()` .

Usipov:
I have added support for cyrillic letters in my fork of this repo.
This was done as a proof of concept only.
I will probably refactor this code to make support for different languages more scalable and then merge the fork back. I just have no time for this at the moment

You may take a look at this PR: https://github.com/Usipov/EarlGrey/pull/1/files

tirodkar:
That's super awesome. I was planning to do the following to add support:

* Allow custom gShiftLabels to be set, with an override option.
* Allow custom modifier key identifiers - space, delete return etc.
* Allow custom more, numbers & more, letters & Next Keyboard key identifiers.
* Allow the ability for checking capitalization with the custom language.

Would that satisfy all the requirements?

Usipov:
@tirodkar, no, that's wouldn't
some characters require long tap and a following swipe gesture
![simulator screen shot 9 aug 2017 10 17 16](https://user-images.githubusercontent.com/1080247/29109520-eef54df4-7ceb-11e7-81c5-bda395381fa7.png)

PR #640, opened 2017-10-07
Fix missing fields in assertion error (#639)

artyom-razinov:
There are more fields in error in logs of failure handler than in inout error parameter.

Previous code replaces fields (so some are lost) with only `kErrorDetailActionNameKey` and `kErrorDetailElementMatcherKey`, new code only modifies `kErrorDetailElementMatcherKey`.

See issue: https://github.com/google/EarlGrey/issues/639


tirodkar:
This looks great. It would be great if you could add this to grey_handleFailureOfAssertion as well. Thanks a lot!

artyom-razinov:
Ok! 👌

khandpur:
Hey @artyom-razinov, just checking if you still plan on working on this in future.

tirodkar:
@artyom-razinov we could make any changes and add them to the cl as well to preserve your commit, you'd need to allow contributions to it from your side though.

artyom-razinov:
@tirodkar 

>  It would be great if you could add this to grey_handleFailureOfAssertion as well. 

I've just added it. Sorry for the delay, in my company we planned to gradually get rid of our fork of EarlGrey, but it is not a high priority task.

We also have support of Russian keyboard, however the code is not ready to be pushed to EarlGrey, because it is not scalable for every other keyboard. Is there any progress on supporting different keyboard layouts already? (https://github.com/google/EarlGrey/issues/31)

If not I'll make a PR maybe this month.

tirodkar:
A PR for that would be great. You can use grey_replaceText() to type in different languages till then.

PR #646, opened 2017-10-18
Make EarlGrey and Functional tests work for deployment target <= 10.0.

mbaxley:
Remove use of APIS deprecated on iOS 10 deployment target:
Replace use of OSAtomic* APIs with methods from stdatomic.h.
Use notifyWhenInteractionChangesUsingBlock as a replacement for
notifyWhenInteractionEndsUsingBlock on iOS 10.0 and higher.

Disable tests that fail when deployment target is 9.0 or higher:
Disable failing tests that don't work when deployment target is 9.0 or
higher.
testSynchronizationWorksWithNSURLConnection relies upon a deprecated API,
so this test doesn't need to be run on iOS 9.0 or higher.
testStyledAlertView induces an alert on completion, which causes other tests
to fail, if the deployment target is 9.0 or higher.

### DIFF
--- a/EarlGrey/Additions/UIViewController+GREYAdditions.m
+++ b/EarlGrey/Additions/UIViewController+GREYAdditions.m
@@ -138,17 +138,18 @@ __attribute__((constructor)) static void initialize(void) {
       // Interactive transitions can cancel and cause imbalance of will and did calls.
       id<UIViewControllerTransitionCoordinator> coordinator = [self transitionCoordinator];
       if (coordinator && [coordinator initiallyInteractive]) {
-        id block = ^(id<UIViewControllerTransitionCoordinatorContext> context) {
-          if ([context isCancelled]) {
-            id object =
-                objc_getAssociatedObject(self, @selector(greyswizzled_viewWillAppear:));
-            UNTRACK_STATE_FOR_OBJECT(kGREYPendingViewsToAppear, object);
-          }
-        };
-#if !defined(__IPHONE_10_0) || __IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_10_0
-        [coordinator notifyWhenInteractionEndsUsingBlock:block];
+        void (^contextBlock)(id<UIViewControllerTransitionCoordinatorContext>) =
+            ^(id<UIViewControllerTransitionCoordinatorContext> context) {
+              if ([context isCancelled]) {
+                id object =
+                    objc_getAssociatedObject(self, @selector(greyswizzled_viewWillAppear:));
+                UNTRACK_STATE_FOR_OBJECT(kGREYPendingViewsToAppear, object);
+              }
+            };
+#if !defined(__IPHONE_10_0) || (__IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_10_0)
+        [coordinator notifyWhenInteractionEndsUsingBlock:contextBlock];
 #else
-        [coordinator notifyWhenInteractionChangesUsingBlock:block];
+        [coordinator notifyWhenInteractionChangesUsingBlock:contextBlock];
 #endif
       }
 
@@ -183,17 +184,18 @@ __attribute__((constructor)) static void initialize(void) {
     // Interactive transitions can cancel and cause imbalance of will and did calls.
     id<UIViewControllerTransitionCoordinator> coordinator = [self transitionCoordinator];
     if (coordinator && [coordinator initiallyInteractive]) {
-      id block = ^(id<UIViewControllerTransitionCoordinatorContext> context) {
-        if ([context isCancelled]) {
-          GREYAppStateTrackerObject *object =
-              objc_getAssociatedObject(self, @selector(greyswizzled_viewWillAppear:));
-          UNTRACK_STATE_FOR_OBJECT(kGREYPendingViewsToDisappear, object);
-        }
-      };
-#if !defined(__IPHONE_10_0) || __IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_10_0
-      [coordinator notifyWhenInteractionEndsUsingBlock:block];
+      void (^contextBlock)(id<UIViewControllerTransitionCoordinatorContext>) =
+          ^(id<UIViewControllerTransitionCoordinatorContext> context) {
+            if ([context isCancelled]) {
+              GREYAppStateTrackerObject *object =
+                  objc_getAssociatedObject(self, @selector(greyswizzled_viewWillAppear:));
+              UNTRACK_STATE_FOR_OBJECT(kGREYPendingViewsToDisappear, object);
+            }
+          };
+#if !defined(__IPHONE_10_0) || (__IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_10_0)
+      [coordinator notifyWhenInteractionEndsUsingBlock:contextBlock];
 #else
-      [coordinator notifyWhenInteractionChangesUsingBlock:block];
+      [coordinator notifyWhenInteractionChangesUsingBlock:contextBlock];
 #endif
     }
 
--- a/EarlGrey/Additions/UIViewController+GREYAdditions.m
+++ b/EarlGrey/Additions/UIViewController+GREYAdditions.m
@@ -138,14 +138,18 @@ __attribute__((constructor)) static void initialize(void) {
       // Interactive transitions can cancel and cause imbalance of will and did calls.
       id<UIViewControllerTransitionCoordinator> coordinator = [self transitionCoordinator];
       if (coordinator && [coordinator initiallyInteractive]) {
-        [coordinator notifyWhenInteractionEndsUsingBlock:
-         ^(id<UIViewControllerTransitionCoordinatorContext> context) {
-           if ([context isCancelled]) {
-             id object =
-                 objc_getAssociatedObject(self, @selector(greyswizzled_viewWillAppear:));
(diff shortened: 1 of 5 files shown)


mbaxley:
This patch is to address issue 643.

I saw a few examples of conditionally disabled tests so I followed that, even though these are disabled at compile time.

I ran it locally on iOS 11 with deployment set to 8.0 (and 10.0) and both passed.

Let me know what you think!

EarlGrey/Additions/UIViewController+GREYAdditions.m
```diff
@@ -138,14 +138,18 @@ - (void)greyswizzled_viewWillAppear:(BOOL)animated {
...
-           }
-         }];
+        id block = ^(id<UIViewControllerTransitionCoordinatorContext> context) {
+          if ([context isCancelled]) {
+            id object =
+                objc_getAssociatedObject(self, @selector(greyswizzled_viewWillAppear:));
+            UNTRACK_STATE_FOR_OBJECT(kGREYPendingViewsToAppear, object);
+          }
+        };
+#if !defined(__IPHONE_10_0) || __IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_10_0
```
tirodkar: You could also use iOS10_OR_ABOVE() from the defines.

https://github.com/google/EarlGrey/blob/master/EarlGrey/Common/GREYDefines.h#L35

tirodkar: Same throughout.

mbaxley: Will that work? I thought that would just check the current runtime, and not the deployment target being used.

tirodkar: Oh, you're right on that. Just realized we don't have any defines for compile time checks. 

If repeated, feel free to add them in GREYDefines or so, otherwise, this looks good to me.

EarlGrey/Additions/UIViewController+GREYAdditions.m
```diff
@@ -179,14 +183,18 @@ - (void)greyswizzled_viewWillDisappear:(BOOL)animated {
...
+          GREYAppStateTrackerObject *object =
+              objc_getAssociatedObject(self, @selector(greyswizzled_viewWillAppear:));
+          UNTRACK_STATE_FOR_OBJECT(kGREYPendingViewsToDisappear, object);
+        }
+      };
+#if !defined(__IPHONE_10_0) || __IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_10_0
+      [coordinator notifyWhenInteractionEndsUsingBlock:block];
+#else
+      [coordinator notifyWhenInteractionChangesUsingBlock:block];
+#endif
```
e-noyau: Sorry for the late comment, but this could also be written as:

      if @available(iOS 10, *) {
        [coordinator notifyWhenInteractionChangesUsingBlock:block];
      }
    #if !defined(__IPHONE_10_0) || __IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_10_0
      else {
        [coordinator notifyWhenInteractionEndsUsingBlock:block];
      }
    #endif

This would have the advantage of invoking the iOS10 API on ios 10 and above even if the code is compiled for a minimum target or 8 or 9.

mbaxley: @e-noyau  -
I went with the current implementation primarily to make the code a little more readable (my opinion). This also will maintain existing behavior for current iOS 10 users who are using 8.0 as the deployment target.

I understand the advantage of your suggestion, I'll let @tirodkar or @sskhandp decide which they prefer. I'm happy to change it.

Thanks!

mbaxley:
@sskhandp 
I ran this patch locally on the following configurations, and everything passed (except one case that also fails without my patch)

device - OS - deployment target
iPhone 8 Plus - iOS 11 - 10.0
iPhone 8 Plus - iOS 11 - 8.0
iPad Air - iOS 9 - 8.0
iPad - iOS 8.1 - 8.0
iPhone 5 - iOS 8.1 - 8.0 (this failed one test, which also fails without my patch: FTRLocalUIWebViewTest testTextFieldInteraction)

Let me know if there is anything else you'd like me to run.

EarlGrey/Additions/UIViewController+GREYAdditions.m
```diff
@@ -138,14 +138,18 @@ - (void)greyswizzled_viewWillAppear:(BOOL)animated {
...
       if (coordinator && [coordinator initiallyInteractive]) {
-        [coordinator notifyWhenInteractionEndsUsingBlock:
-         ^(id<UIViewControllerTransitionCoordinatorContext> context) {
-           if ([context isCancelled]) {
-             id object =
-                 objc_getAssociatedObject(self, @selector(greyswizzled_viewWillAppear:));
-             UNTRACK_STATE_FOR_OBJECT(kGREYPendingViewsToAppear, object);
-           }
-         }];
+        id block = ^(id<UIViewControllerTransitionCoordinatorContext> context) {
```
khandpur: can you use a stronger type than id for block?

EarlGrey/Additions/UIViewController+GREYAdditions.m
khandpur: TBH - either is fine with me but we haven't use availability macros like @e-noyau anywhere in the codebase before so slightly leaning towards @mbaxley's solution.

Tests/FunctionalTests/Sources/FTRAlertViewTest.m
```diff
@@ -59,6 +59,11 @@ - (void)testAlertViewChain {
 }
 
 - (void)testStyledAlertView {
+  // TODO: Fix this test when the deployment target is 9.0 or greater.
+  // Link: https://github.com/google/EarlGrey/issues/645
+#if defined(__IPHONE_9_0) && (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_9_0)
```
khandpur: You could just wrap the whole test method in that #if defined?

Tests/FunctionalTests/Sources/FTRNetworkTest.m
```diff
@@ -29,6 +29,10 @@ - (void)setUp {
 }
 
 - (void)testSynchronizationWorksWithNSURLConnection {
+// This test uses APIs that don't exist when deployment target is 9.0 or greater.
+#if defined(__IPHONE_9_0) && (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_9_0)
```
khandpur: ditto

Tests/FunctionalTests/TestRig/Sources/FTRNetworkTestViewController.m
```diff
@@ -68,11 +68,13 @@ - (void)verifyReceivedData:(NSData *)data {
   }
 }
 
+#if !defined(__IPHONE_9_0) || (__IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_9_0)
```
khandpur: so what's the way to do iboutlets on 9 and above?

mbaxley: Could you clarify this comment? I think IBOutlets are fine in iOS 9, it's the contents of the method that don't work on iOS 9.

So I think the code is incorrect, and I can move the #if inside the method. The button uses NSURLConnection, which is deprecated on iOS 9. So I don't know if we should replace the contents with something else. The replacement is NSURLSession, which are tested in other methods.

Let me know if you want me to move the #if into the method (and add a comment explaining what is going on). If I'm missing the point or you have other suggestions, let me know.

khandpur: I misread your cl description. If the problem is that NSURLConnection is deprecated and causing compilation failure (deprecation turns to error) when you bump the deployment target then I think we should silence that failure and still run it because a user may have done that and we want to make sure EarlGrey can properly synchronize with it. We use the availability macros only for cases where API is completely removed - in this case it's deprecated not removed.

EarlGrey/Additions/UIViewController+GREYAdditions.m
mbaxley: Done.

Tests/FunctionalTests/Sources/FTRAlertViewTest.m
mbaxley: Done

Tests/FunctionalTests/Sources/FTRNetworkTest.m
mbaxley: I removed this ifdef, by suppressing the deprecation error, this test now works.

Tests/FunctionalTests/TestRig/Sources/FTRNetworkTestViewController.m
mbaxley: I suppressed the deprecation error for this call.

mbaxley:
@sskhandp 

I uploaded a second patch to address the comments. I talked to some folks, and they saw nothing wrong with suppressing the deprecation error for the test code.

Let me know what you think.

khandpur:
Thanks for checking with them. LGTM from my side.

tirodkar:
Saw some failures in the CI. They're probably unrelated. I'll wait for the CI before merging. Thanks.

Issue #650, opened 2017-10-23
EarlGreyExample Swift tests compile error on Xcode 9.0.1


bootstraponline:
Repro steps:
1. `git clone git@github.com:google/EarlGrey.git`
0. `cd EarlGrey/Demo/EarlGreyExample/`
0. `pod install`
0. `open EarlGreyExample.xcworkspace/`
0. Run the `testBasicSelectionAndAction`

Error:
![earlgrey_example_2](https://user-images.githubusercontent.com/1173057/31903405-fb10510c-b7f5-11e7-87ce-09c4599ae6b1.png)

I'm seeing this error on a mac using `Xcode 9.0.1`. It doesn't reproduce on other machines. Any idea what's wrong?

bootstraponline:
![earlgrey_example](https://user-images.githubusercontent.com/1173057/31903977-e201937c-b7f7-11e7-83f3-a2d7f6250866.png)

tirodkar:
I recall the error with `kGREYFailureHandlerKey` from issues where EarlGrey was not correctly linked to the target where EarlGrey.swift was present. Can you check if this is the case of if EarlGrey.swift might be replicated in the wrong target?

bootstraponline:
I'll check.

bootstraponline:
I reinstalled Ruby with no luck. Same error. The xcode `EarlGrey.swift` and `EarlGrey.framework` look correct to me. Any other ideas?

<img width="949" alt="earglrey_debug_1" src="https://user-images.githubusercontent.com/1173057/31909481-4a117f44-b808-11e7-923c-c30af557b025.png">
<img width="699" alt="earlgrey_debug_2" src="https://user-images.githubusercontent.com/1173057/31909485-4b2d8738-b808-11e7-972b-dcb1090b9a7e.png">

bootstraponline:
Here's the project. Run `testBasicSelectionAndAction` in the `EarlGreyExampleSwiftTests` target. Repros for me using the zip.

https://drive.google.com/file/d/0Bw73FArizy7GMzdwQnVuU0tZckk/view?usp=sharing

bootstraponline:
I think this is the problem:

```
PODS:
  - EarlGrey (1.4.0)

DEPENDENCIES:
  - EarlGrey

SPEC CHECKSUMS:
  EarlGrey: 763d04569f96d910cf17e4fe5aa0d3842e07c0b4

[...]

PR #651, opened 2017-10-23
pod install --repo-update is required

bootstraponline:
Fix https://github.com/google/EarlGrey/issues/650


Demo/EarlGreyExample/README.md
```diff
@@ -11,7 +11,7 @@ different functionalities provide by EarlGrey, written in Objective-C and Swift.
 
 ### CocoaPods
 
-Execute `gem install earlgrey` then `pod install` from the directory where you have downloaded
+Execute `gem install earlgrey` then `pod install --repo-update` from the directory where you have downloaded
 the source of the example, then `open EarlGreyExample.xcworkspace` to open the example in xcode.
```
tirodkar: Won't `--repo-update` update all the repos in the spec irrespective of whether a repo is required for EarlGrey or not? Maybe have this as part of a note?

bootstraponline: Yes, the Example only has one pod though. We'd want that to be up to date. I'll update the readme to clarify.

bootstraponline: Updated. https://github.com/CocoaPods/CocoaPods/issues/5697#issuecomment-236604410 has more background.

PR #660, opened 2017-11-09
Use TIPreferencesController to change the keyboard settings so it wil…

wuhao5:
…l not load TIUserWordsManager, which can cause occasional crashes on iOS 11.0+.


wuhao5:
Running this on Travis iOS matrix setup.

wuhao5:
the flaky is not related to this change, i think we are good here.

PR #667, opened 2017-12-05
Make Javascript evaluation work with WKWebViews 

petaren:
Added a conditional to make executing javascript on a webview work on WKWebViews too. Uses the same API so nobody needs to change their code.


reshmaraghu:
Hello,
Wanted to know if this will fix the issue I am facing and have posted the same here:
https://stackoverflow.com/q/49622556/8433153

Issue #518, opened 2017-05-30
React-native synchronization


bootstraponline:
> Detox takes this into account by monitoring all asynchronous operations that take place in your app from the inside. Detox knows which network requests are currently in-flight. Detox knows how busy the React Native bridge is. Tests are automatically synchronized to the app and only move forward when the app is idle.
>
> https://github.com/wix/detox/blob/master/docs/Troubleshooting.Flakiness.md

It'd be nice to auto sync with React Native out of the box.

LeoNatan:
[Maintainer of Detox iOS idling resources for RN]

I don’t see an issue in general to upstream our idling resources. Right now, there is a known limitation in some of our own idling resources, where we only support one bridge. For an upstream submission, these need to be ironed out.

Other than that, it’s a question of politics, if Google want to potentially dirty up Earl Grey with Facebook’s custom license.

Let’s see if there is demand to upstream our work.

bootstraponline:
[react-native is BSD](https://github.com/facebook/react-native/blob/master/LICENSE) so I don't see an issue? What code from react-native is being used in the custom IdlingResource? If `NSClassFromString` is used then we might not need the Facebook code at all.

Hopefully other companies automating react-native with EarlGrey will comment on this issue.

LeoNatan:
Are you sure it is not tainted by the custom FB license with controversial patents clause? (It does link and use React.js after all.)

At any rate, indeed Detox is indeed not linked directly with RN; it uses reflection to discover classes and use API, further isolating it from potentially being at odds with the FB license.

bootstraponline:
> Are you sure it is not tainted by the custom FB license with controversial patents clause? (It does link and use React.js after all.)

No, you're confusing an additional patent grant with the actual license the code is released under.

> At any rate, indeed Detox is indeed not linked directly with RN; it uses reflection to discover classes and use API, further isolating it from potentially being at odds with the FB license.

That's perfect, there's no issue then.

LeoNatan:
Can you invite me to the E.G. Slack?

bootstraponline:
Yeah, what is your email? I'm code@bootstraponline.com

bootstraponline:
PR has been opened to upstream the idle resource listeners:
https://github.com/google/EarlGrey/pull/703

LeoNatan:
I am sorry, I will not getting involved in that PR, as per my explanation in the comment I left there. Please do with the PR as you will.

PR #679, opened 2018-02-05
Add Support for Device Shake API

LeoNatan:
Implements #678.
Added a functional test to test device shake.

Note that at the time of cloning the project, numerous unit tests are broken and do not pass on iOS 11.)

PR #703, opened 2018-05-01
Idle resource listeners for React Native

armstrongnate:
This addresses issue #518.

All I did was add the idle resource work from Wix's [Detox](https://github.com/wix/detox/tree/master/detox/ios/Detox) library. It's working great for our React Native apps.

### DIFF
--- a/EarlGrey.xcodeproj/project.pbxproj
+++ b/EarlGrey.xcodeproj/project.pbxproj
@@ -110,6 +110,22 @@
 		7CFDF8EB1DBBD5FC00CAADB3 /* GREYLogger.m in Sources */ = {isa = PBXBuildFile; fileRef = 7CFDF8E51DBBD5FC00CAADB3 /* GREYLogger.m */; };
 		7CFDF8EC1DBBD5FC00CAADB3 /* GREYObjectFormatter.h in Headers */ = {isa = PBXBuildFile; fileRef = 7CFDF8E61DBBD5FC00CAADB3 /* GREYObjectFormatter.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		7CFDF8ED1DBBD5FC00CAADB3 /* GREYObjectFormatter.m in Sources */ = {isa = PBXBuildFile; fileRef = 7CFDF8E71DBBD5FC00CAADB3 /* GREYObjectFormatter.m */; };
+		B1B2244A2097D5E7009B1D9D /* ReactNativeSupport.m in Sources */ = {isa = PBXBuildFile; fileRef = B1B2243A2097D5E6009B1D9D /* ReactNativeSupport.m */; };
+		B1B2244B2097D5E7009B1D9D /* EarlGrey+Detox.h in Headers */ = {isa = PBXBuildFile; fileRef = B1B2243B2097D5E6009B1D9D /* EarlGrey+Detox.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		B1B2244C2097D5E7009B1D9D /* ReactNativeHeaders.h in Headers */ = {isa = PBXBuildFile; fileRef = B1B2243C2097D5E6009B1D9D /* ReactNativeHeaders.h */; };
+		B1B2244D2097D5E7009B1D9D /* WXJSDisplayLinkIdlingResource.m in Sources */ = {isa = PBXBuildFile; fileRef = B1B2243D2097D5E6009B1D9D /* WXJSDisplayLinkIdlingResource.m */; };
+		B1B2244E2097D5E7009B1D9D /* GREYMatchers+Detox.m in Sources */ = {isa = PBXBuildFile; fileRef = B1B2243E2097D5E6009B1D9D /* GREYMatchers+Detox.m */; };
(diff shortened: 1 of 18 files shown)


LeoNatan:
The API as-is is clearly not ready for general purpose Earl Grey, and just dumping them is really not a preferred solution.

For example, some important issues:
* Only support one RN bridge instance—This may be a valid limitation for Detox (for now), but I don't think it's a good idea to have it as is in Earl Grey
* No proper strategy to tackle multiple versions of RN, which tend to break the API—This is very important; we have ideas but I have not had time to implement them as our current solution is good enough™ for Detox—but not for EG
* Not extensible—Currently, if a native module does its own hocus pocus in native, there is no API to let developers extend Detox/idling resources to add support for their stuff. Again, we have ideas for this, but have not had the need to implement
* Some code is very hacky

In addition to the problems above (and others), this is just a code dump straight from wix/Detox. Where is the test plan? RN is a volatile project with breaking changes almost every version. What is the plan in Earl Grey to test new versions? What is the commitment from Google and the project maintainers/collaborators to support new versions and old? We wouldn't mind filling that role if this code does find its way into Earl Grey, but we need to have assurances that we will have quick access. We cannot agree to a PR ping pong that will take [almost two months](https://github.com/google/EarlGrey/pull/679) to be accepted. This is by no means a criticism of the process, only to underline the necessity for the ability to quickly be able to fix issues. Additional factors also play, as internally we must support specific RN versions that are used in our products. Sometimes those version lag behind the officials by a large number of versions. While we do our best to support all modern versions, sometimes we only support the latest version and the internal versions. Having to maintain a support matrix for many multiple versions of RN is not something I can (or want!) to commit to. This is fine for Detox, but since no official policy has been discussed for Earl Grey, I'd assume a much larger matrix would be needed here. Who will support that?

All of this, and we haven't even discussed EGv2 and the huge change in API that that will entail. I have not have even considered what that would mean for the Detox extensions, or how to approach a solution. Since it will be out of process, at least part of it, Detox may not even be able to switch to that, or at least not without a huge work behind the scenes. Whether this work will happen for Detox is an open question, predicated on the aforementioned lack of consideration for v2 so far.

On a personal note, I do not appreciate such a blatant code dump, my name and Wix's plastered in headers and all. Even if in good faith, this is not the way to do it. An [issue](https://github.com/google/EarlGrey/issues/518) was created, and discussion should have continued there, getting our agreement and commitment to cooperate with the task, if we can.

As it stands, I will not be involved in this PR. Do with it as you like.

khandpur:
I don't think we can accept a code dump from Detox codebase as-is without proper approvals. I will post more information once I have some clarity on the process.

bootstraponline:
@LeoNatan Thanks for your feedback. The issue about upstreaming the idle resource listeners has been open for 11 months, 1 day without any comment from the EarlGrey team. I see this pull request as a way to move the conversation forward. I agree about not merging it as-is.

I think you've highlighted some real issues with the EarlGrey project. I'll be meeting with the team next week. I'm hoping for some good news about react-native support.

The current option for teams needing react native support with EarlGrey is to code dump from the detox repo which isn't ideal. One solution would b moving the code into a new detox org repo and publishing to Cocoapods so it can be consumed as a dependency directly.

We're required to keep the headers unmodified per the open source license. Pull requests to EarlGrey often remain open for years as reference for technical discussions. I'm hoping we can learn from this use case and figure out a path forward everyone can agree on.

LeoNatan:
Perhaps a better way forward is discussing a plugin infrastructure for EG, where we can release a "DetoxKit" for EG. But again, dumping our code into EG is not the right way to move the discussion. In my opinion, the original issue was kept quiet on Google/collaborator's part because it was left open for us to first finish a discussion that we never had. I don't see how Google can move that conversation anywhere when my team has not even considered any work or possible work on such a conversion, let alone discuss with managers, etc. This is a very big undertaking.

As I mentioned, personally, I don't like that this dump was done without any notice by OP. Even if the license allows it, I don't think it advances anything.

bootstraponline:
> I don't like that this dump was done without any notice by OP. Even if the license allows it, I don't think it advances anything.

I had pinged you a few times on Slack in the past without getting a response. I think this was around the time Appium was looking into a detox based driver. I don't think there's any risk of the PR getting merged, so no worries there. 

The vision is that wix & anyone else testing react native apps on EarlGrey could collaborate on a shared framework that provides idle resource listener support.

Here's the conversation I hoped to have based on the code in the PR:

- Is this feature appropriate for upstream? If so, does Google have an interest in maintaining it?
- How will this code adapt to the new EarlGrey v2 API?
- How do we dramatically improve the pull request response time & build stability so companies can collaborate in the EarlGrey repo?

That's difficult for me to do by simply pointing at the giant detox repo, of which only a small part will be considered for upstreaming. I'm depending on this code for my automation efforts and there's a lot of risk and uncertainty without answers to these questions.

Wix is doing all the hard work so it wasn't my intent to cause unhappiness. Ideally we'd be able to collaborate since we're sharing a common automation approach.

bootstraponline:
Update: I met with the team, after the EarlGrey 2 release is out we'll discuss:
- Evaluating new CI options to improve the feedback loop on PRs. 
- Defining a minimum turn around time to get PRs merged.
- Migrating the detox EarlGrey changes upstream so that EarlGrey doesn't break detox.

bootstraponline:
We've decided to invest in building native Swift apps. As a result, there's no longer a business need for us to upstream react native support.

I think we can close this pull request.

Issue #704, opened 2018-05-04
UIWebViews are not detected as idle when displaying PDFS


keefertaylor:
Per Apple's documentation: 

> "UIWebView objects can be used to display other content types, such as Keynote, PDF, and Pages documents. For the best rendering of plain and rich text in your app, however, you should use UITextView instead." 

(Supported File Formats, https://developer.apple.com/documentation/uikit/uiwebview)

EarlGrey's current implementation of UIWebViewIdlingResource uses a11y elements in the browser view hierarchy to determine idle state. In PDF and other document types, the browser view is not use.  Instead, content is populated inside an internal PDFView property. See: https://github.com/google/EarlGrey/blob/master/EarlGrey/Synchronization/GREYUIWebViewIdlingResource.m#L143

Since EarlGrey waits for content to be rendered inside the browser view, UIWebViews displaying PDFs are never displayed as idle.

PR #705, opened 2018-05-04
Support PDF Display in UIWebViews

keefertaylor:
#704 
In a webview with HTML content, there is a valid view at _webView._internal.browserView, but the view is always empty. This causes EG's UIWebView Idling Resource to assume the webview is waiting to render.

In a webview with a PDF, there is a view at _webView._internal.pdfHandler.pdfView. This view *only* exists after the PDF is rendered (AFAICT) and does not exist if the webview is rendering something other than a PDF.

To fix, check if there is a rendered PDF in the webview before performing HTML rendering checks.


EarlGrey/Synchronization/GREYUIWebViewIdlingResource.m
```diff
@@ -137,6 +137,11 @@ - (BOOL)isIdleNow {
     }
   }
 
+  // UIWebViews may be used to display PDFs in addition to HTML based content. Test if there's a PDF
+  // view populated in UIWebView's hierarchy.
+  BOOL webViewIsDisplayingPDF = ([[[_webView valueForKey:@"_internal"] valueForKey:@"pdfHandler"]
```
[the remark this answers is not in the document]
keefertaylor: > Do we have to do something with the bool?
Yup, need to return that we are idle.

> Can we add a test case with a PDF to prevent regressions & prove the issue is fixed?
Sure, let me put something together.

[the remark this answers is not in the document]
keefertaylor: I wired up a basic test case that shows that opening a PDF will cause the app to never reach an idle state. PTAL.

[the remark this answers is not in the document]
bootstraponline: Nice, looks good to me. Someone on the EarlGrey will need to approve before the PR is merged.

keefertaylor:
Adding a comment to get CLABot to re-run.

EarlGrey.xcodeproj/xcshareddata/xcschemes/EarlGrey.xcscheme
```diff
@@ -26,16 +26,37 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      language = ""
```
khandpur: Are these changes intentional or just a side-effect of opening in a different Xcode?

EarlGrey/Synchronization/GREYUIWebViewIdlingResource.m
```diff
@@ -137,6 +137,14 @@ - (BOOL)isIdleNow {
...
+                                     valueForKey:@"pdfView"] == nil);
+  if (webViewIsDisplayingPDF) {
```
khandpur: It's not clear to me what's throwing off EarlGrey sync logic.

@tirodkar Can you please see if it's something we can fix in the core-logic rather than short circuiting the entire method when PDFs are being loaded?

Tests/FunctionalTests/Sources/FTRUIWebViewTest.m
```diff
@@ -108,6 +108,13 @@ - (void)DISABLED_testSynchronizationWhenSwitchingBetweenLoadingMethods {
       assertWithMatcher:grey_sufficientlyVisible()];
 }
 
+- (void)testPDFIdlesCorrectly {
```
khandpur: I'd recommend splitting this up with local only and network-only scenarios (for an internal use-case)

Tests/FunctionalTests/TestRig/Sources/FTRWebViewController.m
```diff
@@ -77,6 +84,12 @@ - (IBAction)userDidTapLoadHTMLUsingLoadHTMLString {
   _webView.scrollView.bounces = _webViewBounceSwitch.isOn;
 }
 
+- (IBAction)userDidTapLoadPDF {
+  NSData *pdfData = [NSData dataWithContentsOfURL:[[self class] urlToTestPDF]];
+  [_webView loadData:pdfData MIMEType:@"application/pdf" textEncodingName:@"utf-8"
+   baseURL:[NSURL URLWithString: @"about:blank"]];
```
khandpur: formatting seems off.

tirodkar:
Hey @keefertaylor, we're getting this change in as part of our internal to external code sync with any changes that were requested here. I'll close this for now.

PR #708, opened 2018-05-22
Look for groups matching the test target name outside of the main group

weibel:
In our project we have an intricate group structure and at some point the group containing our specs moved into a subfolder in the hierarchy. 

When we run pod install after adding EarlGrey it will fail with the error

```
RuntimeError - Test target group not found!
/Users/kasper/.rvm/gems/ruby-2.3.1/bundler/gems/EarlGrey-35532c4e95c6/gem/lib/earlgrey/configure_earlgrey.rb:369:in `copy_swift_files'
etc.
```

After looking through the code the culprit seemed to be the way the target names and group names are mathed inside configure_earlgrey.rb in copy_swift_files. In ```test_target_group = project_test_targets.find { |g| g.display_name == target.name }``` only the root groups are looked at, and it did not work for our project.

To solve the problem we have added a new method which will recurse through the group hierarchy until it finds a matching group name.

### DIFF
--- a/gem/lib/earlgrey/configure_earlgrey.rb
+++ b/gem/lib/earlgrey/configure_earlgrey.rb
@@ -366,7 +366,7 @@ module EarlGrey
       project_test_targets = project.main_group.children
       test_target_group = recursively_find_group_with_name(project_test_targets, target.name)
 
-      raise "Test target group not found! #{test_target_group}" unless test_target_group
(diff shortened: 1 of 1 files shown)


tirodkar:
Thanks for looking into this @weibel. Apologies for not getting to it faster. Can you add a simple test for it in our rspec as well?

niamh-power:
@weibel Any updates on this? We've come across the same issue :(

Issue #709, opened 2018-05-22
Earl Grey swizzling fails when proxy objects are used.


LeoNatan:
I've seen this issue with projects using [TrustKit](https://github.com/datatheorem/TrustKit) and [Detox](https://github.com/wix/detox), but this is a general problem of Earl Grey.

Fox example, in NSURLSession+GREYAdditions.m, `greyswizzled_dataTaskWithRequest:completionHandler:`, the code checks if `[delegate respondsToSelector:originalSel]`, but later, `class_getInstanceMethod` returns `NULL` because the `delegate` instance is actually a proxy which implements `respondsToSelector:` and `forwardingTargetForSelector:`.

I will submit a PR with a special case fix for the case above, but consider refactoring any swizzling mechanisms taking into account `forwardingTargetForSelector:` at the very least.

werediver:
I can also confirm that the swizzling applied to `XCTestCase` is quite rough, especially the class-/isa-swizzling applied to `XCTestCase.invocation`. It would be much better to put all the EarlGrey-specific features in some `GREYTestCase`, a subclass of `XCTestCase` (and use as little dynamic alterations as possible), than to screw the default one.

The class-/isa-swizzling on the `XCTestCase.invocation` even does not check the actual type of the invocation, so a proxy object cannot be used there (whereas normally it would work perfectly). Using a wrapping proxy would be a more universal solution for EarlGrey (it would work with `NSInvocation` subclasses [if any; though it's not actually "subclassable"], with proxy objects, whatever).

I have used a per-test (per-XCTestCase-method) tagging for flexible test filtering/grouping. The implementation is quite typical, no dynamic alterations, etc. But EarlGrey is quite incompatible with custom `XCTestCase` subclasses 🙁

rumax:
Is this going to be ever solved? Any expectations?

werediver:
@rumax I believe the EarlGrey team is focused on v2.x nowadays and won't put a lot of effort maintaining v1.x.

This problem is not a tiny bug, though. It's a framework design issue and "fixing" it would require significant effort.

LeoNatan:
The same issue exists in v2, I think. This part of the framework has not changed.

tirodkar:
How is the proxy referred to? I believe if there was a way that the swizzling code could know there is a proxy then we could track it correctly. Alternatively, the proxy object could itself be turned into an idling resource.

LeoNatan:
[...]

PR #710, opened 2018-05-22
Fix TrustKit swizzling issues in Earl Grey

LeoNatan:
As explained in #709, there is an issue with Earl Grey swizzling and proxy objects.
This PR fixes specifically one of these issues when TrustKit is integrated in the host project and swizzles in TrustKit are enabled.

The solution is to attempt getting the forwarding target class, if any, or use the actual class of the delegate as before.

### DIFF
--- a/EarlGrey/Additions/NSURLSession+GREYAdditions.m
+++ b/EarlGrey/Additions/NSURLSession+GREYAdditions.m
@@ -18,6 +18,7 @@
 
 #include <objc/runtime.h>
 
+#import "Additions/NSURL+GREYAdditions.h"
 #import "Additions/__NSCFLocalDataTask_GREYAdditions.h"
 #import "Additions/NSURL+GREYAdditions.h"
 #import "Common/GREYFatalAsserts.h"
@@ -28,9 +29,7 @@
 /**
  *  Type of the handlers used as NSURLSessionTask's completion blocks.
  */
-typedef void (^GREYTaskCompletionBlock)(NSData *data,
-                                        NSURLResponse *response,
-                                        NSError *error);
+typedef void (^GREYTaskCompletionBlock)(NSData *data, NSURLResponse *response, NSError *error);
 
 @implementation NSURLSession (GREYAdditions)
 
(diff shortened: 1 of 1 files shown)


EarlGrey/Additions/NSURLSession+GREYAdditions.m
```diff
@@ -62,6 +62,10 @@ - (NSURLSessionDataTask *)greyswizzled_dataTaskWithRequest:(NSURLRequest *)reque
     // request need not be tracked as its completion/failure does not trigger any delegate
     // callbacks.
     if ([delegate respondsToSelector:originalSel]) {
+      // This is to solve issues where there is a proxy delegate instead of the original instance (e.g. TrustKit).
```
khandpur: You'd need a check in line 59 as well otherwise it will check wrong delegate class for existence of swizzled selector. I feel like there's a cleaner way to write up when and how to swizzle logic:
1) change line 59 to check if delegate responds to swizzledSel
2) inside `@synchronized(selfClass)` first check if [delegateClass instancesRespondToSelector:original], if not, get the forwarding target becase according to https://developer.apple.com/documentation/objectivec/nsobject/1418555-instancesrespondtoselector?language=objc "If aSelector messages are forwarded to other objects, instances of the class are able to receive those messages without error even though this method returns NO." so you should be able to leverage that and grab the right class to swizzle. Now if that class forwards it again...well let's just worry about it if we run into that case :)

LeoNatan: Changing 59 is not necessary. It's what the system asks when deciding which protocol methods are implemented. The documentation might be correct in theory regarding the ObjC runtime, but the way Apple handles delegates is by keeping a packed struct of delegate capabilities, and only calls those delegate methods iff the cap flag of the required method is enabled. The cap flags are set once upon delegate setting. You can verify this by overriding `respondsToSelector:` and printing the selectors. You will notice that the method is called in bulk for each delegate method at the time of setting the delegate.

yogeshthanvi:
@khandpur  When we are expecting this to release because its blocker to test projects.

khandpur:
There are some pending comments that require attn. If @LeoNatan is busy, I could take his changes and start a new PR as long as I get an okay from him to do so :)

LeoNatan:
I have commented above. I don't believe there is anything I haven't addressed.

tirodkar:
@yogeshthanvi Do you need this in a CocoaPods release or building from head?

LeoNatan:
We use a git submodule to bring EG as a dependency.

yogeshthanvi:
This is still an issue with us , is this released @tirodkar ? Please find the comments in above for further details

rumax:
Is this going to be ever merged/close soon? Any expectations?

tirodkar:
+ 1 to the request for adding a Unit Test - @LeoNatan

JakeHiggins:
What is the ETA on this merge getting accepted? Is there any outstanding action required?

plozachm:
> What is the ETA on this merge getting accepted? Is there any outstanding action required?
Same question here @tirodkar, would really help for detox on iOS with Trustkit.
Thank you in advance.

popovprodigyai:
Any ETA on this? 🤕

steve-westwood:
Also waiting on this to integrated to fix detox SSL pinning issue +1

khandpur:
I would like to accept this PR but I still don't see my comments addressed which includes adding a test case for the fixed behavior. Once in, we will port it over to EarlGrey 2.0 ourselves.

LeoNatan:
I don’t have the capacity on adding a test case at this time. I can explain how the runtime works if something is not clear in the issue or PR. THANKS

tirodkar:
Updated the PR with changes done internally. PTAL. I'll merge this once the CI finishes.

LeoNatan:
Thanks guys! We will be updating Detox soon.

plozachm:
Thanks a lot guys, very much appreciated!

PR #719, opened 2018-05-30
Update version number in docs and plist for 1.14.0.

ynzhang0509:


ynzhang0509:
Oh I did not notice you have another pr. But does not matter, I also generated the pdf.

Best regards
Yan Zhang

PR #738, opened 2018-07-27
Updated carthage setup

Macarse:


docs/install-and-run.md
```diff
@@ -79,11 +79,11 @@ and that you keep your EarlGrey version synced to the latest.
 
       `brew install carthage`
 
-   2. Specify the version of EarlGrey to use in Cartfile.private.
```
bootstraponline: Cartfile.private is used for a reason.

https://github.com/Carthage/Carthage/blob/9d1a67f47c7231eb2f1c33a719634ec4281b65c1/Documentation/Artifacts.md#cartfileprivate

> Anything listed in the private Cartfile will not be seen by dependent (parent) projects, which is useful for dependencies that may be important during development, but not when building releases—for example, test frameworks.

Macarse: Thanks. I will create a new PR.

PR #744, opened 2018-08-04
Update docs to fix func name

buhikon:
function name in the install and run document was changed from `testExample` to `testPresenceOfKeyWindow`.


buhikon:
I signed it!

PR #746, opened 2018-08-14
Add script to build universal framework

### LINKED ISSUE #745: Universal Static Library Script
It'd be nice to have a script that generates a universal static library for EarlGrey. Then when EarlGrey is used in CI, the dependency doesn't have to be recompiled.

- https://github.com/AudioKit/AudioKit/blob/master/Frameworks/build_frameworks.sh
- https://github.com/KeepSafe/ObjectiveRocks/blob/frameworks/build_universal_framework.sh
- http://devcenter.bitrise.io/docs/should-i-commit-my-dependencies-into-my-repository

bootstraponline:
Fix #745

### DIFF
--- a/Scripts/.gitignore
+++ b/Scripts/.gitignore
@@ -1,0 +1,2 @@
+tmp/
+universal/
--- a/Scripts/build_universal_framework.sh
+++ b/Scripts/build_universal_framework.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+rm -rf tmp/ universal/
+
+set -euo pipefail
+
+if ! [ -x "$(command -v xcpretty)" ]; then
+  gem install xcpretty
+fi
+
+SCHEME="EarlGrey"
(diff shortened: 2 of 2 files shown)


tirodkar:
If you add this to `EarlGrey.xcodeproj`, will everything work fine?

bootstraponline:
Would you prefer this be part of the xcode project? It's possible to add a shell script.

The problem is it'll slow down the builds for users that don't need a universal framework. Typically the pattern I've seen is to add an external export script.

tirodkar:
That sounds good. LG from my side.

khandpur:
@tirodkar we have something like this we use internally. Why not open source that and let Matt add his changes to that? That way we use one single script to generate framework that we ship to cocoapod users?

bootstraponline:
That sounds awesome.

bootstraponline:
closing in favor of https://github.com/google/EarlGrey/pull/761

PR #754, opened 2018-09-26
Use Old Build System

st3fan:
This patch adds a WorkspaceSettings.xcsettings file to disable new build system. This will fix the issue where EarlGrey will not successfully build for example in CI with Carthage.


tirodkar:
@st3fan I'm trying to rerun our Unit Tests which keep failing after this change. Once it's fixed, I'll get this in. Might possibly be unrelated errors.

ethanjdiamond:
This is blocking my team as well. Do you have an idea when you'll get to it?

PR #766, opened 2018-10-23
Adding support for WKWebView

jeromelebel:
This patch adds support for WKWebView using grey_javaScriptExecution().


tirodkar:
Cool, is this the method from EG 2 moved to EG 1?

tirodkar:
Merging since the failures are all known.

PR #780, opened 2018-12-21
Update RVM to 2.3.0 as it's failing to update

khandpur:
https://travis-ci.org/google/EarlGrey/jobs/471151306:
```
Updating rubygems-update
ERROR:  Error installing rubygems-update:
	rubygems-update requires Ruby version >= 2.3.0.
ERROR:  While executing gem ... (NoMethodError)
    undefined method `version' for nil:NilClass
The command "gem update --system" failed and exited with 1 during .
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ osx_image: xcode8.3
 matrix:
   global:
     rvm:
-      - 2.2.5
+      - 2.3.0
   include:
     - env:
         - DESTINATION="OS=8.1,name=iPhone 4s" SDK=iphonesimulator10.3 TYPE="FUNCTIONAL"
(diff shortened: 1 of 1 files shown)


wuhao5:
The ruby issue should be fixed, not sure about the other failure, shouldn't be related?

khandpur:
Right. The unit test failures are known. Waiting on FTR to complete to see what other failures there are.

tirodkar:
Yeah, we've seen the other failure before with the Unit tests. There seems to be a blacklisting URL issue that we've seen on Travis but I wasn't able to reproduce locally. This cl is good to go.

khandpur:
Functional failures are also known failures. This is ready to merge.

Issue #781, opened 2018-12-26
EarlGrey 2.0 | Exception when checking for Accessibility ID 


mitchell-asb:
I have an accessibilityIdentifier that is a number and its failing on the `isEqualToString` selector as `__NSCFNumber` doesnt have that selector. I will submit a pr adding a check if the class instance supports the selector `isEqualToString` unless anyone else has a better idea (maybe a cast from number to string?)
```
Uncaught exception in Application: -[__NSCFNumber isEqualToString:]: unrecognized selector sent to instance 0xd589f038d3507a35
0   CoreFoundation                      0x000000010ac0b1bb __exceptionPreprocess + 331
1   libobjc.A.dylib                     0x000000010980a735 objc_exception_throw + 48
2   CoreFoundation                      0x000000010ac29f44 -[NSObject(NSObject) doesNotRecognizeSelector:] + 132
3   CoreFoundation                      0x000000010ac0fed6 ___forwarding___ + 1446
4   CoreFoundation                      0x000000010ac11da8 _CF_forwarding_prep_0 + 120
5   AppFramework                        0x0000000105548a24 __42+[GREYMatchers matcherForAccessibilityID:]_block_invoke + 388
6   AppFramework                        0x0000000105536868 -[GREYElementMatcherBlock matches:] + 72
7   AppFramework                        0x0000000105523425 -[GREYBaseMatcher matches:describingMismatchTo:] + 85
8   AppFramework                        0x0000000105558377 -[GREYAllOf matches:describingMismatchTo:] + 471
9   AppFramework                        0x0000000105558127 -[GREYAllOf matches:] + 167
10  AppFramework                        0x0000000105560c3c -[GREYElementFinder elementsMatchedInProvider:] + 1260
11  AppFramework                        0x0000000105565ed2 __78-[GREYElementInteraction matchElementsWithTimeout:syncBeforeMatch:matchBlock:]_block_invoke + 322
12  AppFramework                        0x00000001055139b4 __59-[GREYUIThreadExecutor executeSyncWithTimeout:block:error:]_block_invoke + 68
13  AppFramework                        0x00000001054fe4ca __98-[GREYRunLoopSpinner grey_drainRunLoopInActiveModeAndCheckCondition:explicitRunLoopDrain:forTime:]_block_invoke + 250
[...]

PR #782, opened 2018-12-26
Resolve invalid selector `isEqualToString` being called #781

mitchell-johnson:
Resolve issue where invalid selector `isEqualToString` being called on a number #781

### DIFF
--- a/AppFramework/Matcher/GREYMatchers.m
+++ b/AppFramework/Matcher/GREYMatchers.m
@@ -151,7 +151,11 @@ static Class gEDOObjectClass;
     if (element.accessibilityIdentifier == accessibilityID) {
       return YES;
     }
-    return [element.accessibilityIdentifier isEqualToString:accessibilityID];
+
+    if( [element.accessibilityIdentifier respondsToSelector:@selector(isEqualToString:)]){
+      return [element.accessibilityIdentifier isEqualToString:accessibilityID];
+    }
+    return NO;
   };
   GREYDescribeToBlock describe = ^void(id<GREYDescription> description) {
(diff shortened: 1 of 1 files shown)


tirodkar:
https://developer.apple.com/documentation/uikit/uiaccessibilityidentification/1623132-accessibilityidentifier 

Accessibility Identifier is always an NSString, so it has to implement `isEqualToString:`. Can you provide an example of what you're facing so we can see just how this might be able to fix the issue?

Also, you need to sign our CLA.

mitchell-johnson:
I discovered my issue. I had accidentally added an `accessibilityIdentifier` as a BOOL in an element on the affected page. although it's not valid to have a non string `accessibilityIdentifier` there is nothing to stop you doing it accidentally or deliberately. So I wonder if it's worth keeping this check?

mitchell-asb:
Yes I am ok with this PR

AppFramework/Matcher/GREYMatchers.m
```diff
@@ -151,7 +151,11 @@ + (void)initialize {
     if (element.accessibilityIdentifier == accessibilityID) {
       return YES;
     }
-    return [element.accessibilityIdentifier isEqualToString:accessibilityID];
+
```
wuhao5: how about `[accessibilityID isEqualToString:element.accessibilityIdentifier]`?

mitchell-asb: This is actually a really good idea.

mitchell-johnson: This is actually a really good idea.

mitchell-johnson:
@wuhao5 Thats a great idea and adds no extra overhead

PR #788, opened 2019-01-18
Synch in Code for EarlGrey 2.0 for Internal Changes. 

tirodkar:
Chief Changes:

* Error Logging Updates.
* Matchers / Assertions supported from the Test Side.
* Visibility Checker Updates to be more relaxed.
* Calls from the Test Side now have their own queue to allow better IPC.
* EG waits 30 seconds for the app to be launched before starting its own interactions.


khandpur:
The failing bitrise run detail logs show the tests ran to completion and also shows this error:
❌  error: Signing for "FunctionalTestRig" requires a development team. Select a development team in the project editor. (in target 'FunctionalTestRig')

Should this be ignored?

tirodkar:
The bitrise failure on deploy is for an FTL test setup we did which needs signing which isn't available right now.  The Sim tests are all working fine.

PR #806, opened 2019-03-19
Fix warnings

bootstraponline:
eg2_warnings

### DIFF
--- a/Demo/DemoApp/DemoApp.xcodeproj/project.pbxproj
+++ b/Demo/DemoApp/DemoApp.xcodeproj/project.pbxproj
@@ -470,7 +470,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_STYLE = Automatic;
-				DEVELOPMENT_TEAM = B6333T4PXQ;
+				DEVELOPMENT_TEAM = "";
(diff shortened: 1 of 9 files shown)


Demo/DemoApp/DemoApp.xcodeproj/project.pbxproj
```diff
@@ -488,6 +489,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_STYLE = Automatic;
+				DEVELOPMENT_TEAM = B6333T4PXQ;
```
tirodkar: What team are you setting this to? Is this needed?

bootstraponline: I set it to something so I can replace it in the build like this:

> sed -i '' 's/DEVELOPMENT_TEAM = B6333T4PXQ;/DEVELOPMENT_TEAM = 72SA8V3WYL;/g' "$PROJ"

Is there a better value to set it to? If it's absent then it's more painful to update via a shell script.

bootstraponline: Looks like empty string works.

bootstraponline:
The build is passing now. :)

PR #808, opened 2019-03-26
Fix for EG Accessibility break on iOS 12.2 on device.

tirodkar:


snowwolflibo:
@tirodkar 
how is it going? When can I get the correct version from CocoaPods?

tirodkar:
I don't have an eta at the moment. I'll do one last release with the Swift 5 fix as well soon.

PR #838, opened 2019-05-24
Corrected typos in comments, strings, and docs

cbracken:
Non-comment code changes are limited to:
  * local variable rename in `Scripts/rename-ochamcrestIOS.py`.
  * string value change in `Tests/FunctionalTests/Sources/FTRPinchViewTest.m`


cbracken:
Identified using [misspell](https://github.com/client9/misspell) tool while I was doing a pass over our codebase in [Flutter](https://github.com/flutter/flutter). Internal shortID is cbracken@ if any questions.

khandpur:
Thanks Chris!
@tirodkar we should port this over to EG2 as well and enable misspell as a presubmit check if it exists.

PR #958, opened 2019-07-02
Updates AppFramework to compile with a deployment target of 11.0.

rohit-rao:
A UIViewControllerTransitionCoordinator method was renamed in iOS 10.0
and the old method was deprecated.  Calls to this method are now gated
with compile-time guards and runtime checks to ensure that the code can
compile with any deployment target and still run properly on older iOS
versions.

### DIFF
--- a/AppFramework/Additions/UIViewController+GREYApp.m
+++ b/AppFramework/Additions/UIViewController+GREYApp.m
@@ -139,7 +139,13 @@ __attribute__((constructor)) static void initialize(void) {
                 UNTRACK_STATE_FOR_OBJECT(kGREYPendingViewsToAppear, object);
               }
             };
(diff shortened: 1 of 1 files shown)


AppFramework/Additions/UIViewController+GREYApp.m
```diff
@@ -139,7 +139,13 @@ - (void)greyswizzled_viewWillAppear:(BOOL)animated {
                 UNTRACK_STATE_FOR_OBJECT(kGREYPendingViewsToAppear, object);
               }
             };
-        [coordinator notifyWhenInteractionEndsUsingBlock:contextBlock];
+        if (@available(iOS 10.0, *)) {
```
tirodkar: The minimum version is 10, so we don't need this check.

PR #1016, opened 2019-08-16
Fix build-release.sh script

szymon-kazmierczak:
Script was failing due to erroneous backslash.


tirodkar:
Please sign the cla. What is the break that you're seeing?

szymon-kazmierczak:
@googlebot I signed it!

szymon-kazmierczak:
When running the release build, this script is invoked and fails with the 
`'-xcodebuild: error: Unknown build action ' -project'.`

PR #1127, opened 2019-11-28
[FIX] Add GREYTraversalProperties to xcodeproj

Simon-Kaz:
After seeing the following error:

Undefined symbols for architecture x86_64:
  “_OBJC_CLASS_$_GREYTraversalViewProperties”, referenced from:
      objc-class-ref in libUILib.a(GREYTraversalFunctions.o)
I fixed it by added GREYTraversalProperties.h and GREYTraversalProperties.m to the project as they were not there.


tirodkar:
Thank you for the fix @Simon-Kaz! We'll monitor this better in the future.

Issue #1192, opened 2020-02-14
EarlGrey 2.0 | Runtime crash in Demo/EarlGreyExample app


adamhowardprice:
Using Xcode 11.3 / MacOS Catalina 10.15.3

Steps to reproduce:
1. Clone EarlGrey repo, check out `earlgrey2` branch
2. ` sh Scripts/download_deps.sh`
3. `cd Demo/EarlGreyExample`
4.  `pod install`
5. `open EarlGreyExample.xcworkspace` (open in Xcode 11.3)
6. CMD-U to run all tests (or just run any UI test)

Expected Result: The UI tests run
Actual Result: The app crashes at runtime. Error: `libc++abi.dylib: terminating with uncaught exception of type NSException`
Screenshot:

![Screen Shot 2020-02-14 at 5 53 18 PM](https://user-images.githubusercontent.com/602596/74574972-4d8a2380-4f53-11ea-981c-06631a09af72.png)

tirodkar:
Can you make sure the Podfile is pinned to 0.9.0 for eDO?

adamhowardprice:
Podfile.lock:

```
PODS:
  - EarlGreyApp (2.0.0)
  - EarlGreyTest (2.0.0):
    - eDistantObject
  - eDistantObject (0.9.1)

DEPENDENCIES:
  - EarlGreyApp
  - EarlGreyTest

SPEC REPOS:
  trunk:
    - EarlGreyApp
    - EarlGreyTest
    - eDistantObject

SPEC CHECKSUMS:
  EarlGreyApp: e189faf4f55a190ebc4e1c3b2c1d155c2cd2c946
  EarlGreyTest: fe152c2dd32607996f518eb097223b16c94a1806
  eDistantObject: 5da72e81689a31ee7c440f9a6089adf0259cc99e

PODFILE CHECKSUM: 3e3159bde4db479df47b98c15a8f134d8fabe346

COCOAPODS: 1.8.4
```

tirodkar:
Update this file to   - eDistantObject (0.9.0) and then re-do pod install.

adamhowardprice:
That worked. Would it make sense to submit a PR to change that in the Demo Podfile?

tirodkar:
Send a PR for the Podfile.lock. We'll update this once the next EG 2 release happens.

adamhowardprice:
https://github.com/google/EarlGrey/pull/1193

adamhowardprice:
[...]

PR #1193, opened 2020-02-14
Fix demo Podfile to pin eDistantObject to 0.9.0

adamhowardprice:
Resolves https://github.com/google/EarlGrey/issues/1192

### DIFF
--- a/docs/cocoapods-setup.md
+++ b/docs/cocoapods-setup.md
@@ -11,12 +11,15 @@ using EarlGrey 2.0 with.
 Here is an extremely simple Podfile demonstrating integrating an app called
 `App` with EarlGrey 2.0.
 
+__Note:__ You must pin the transitive dependency `eDistantObject` at 0.9.0.
+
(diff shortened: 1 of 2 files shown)


docs/cocoapods-setup.md
```diff
@@ -9,14 +9,16 @@ setup process is fairly straightforward and typical.
 using EarlGrey 2.0 with.
 
 Here is an extremely simple Podfile demonstrating integrating an app called
-`App` with EarlGrey 2.0.
+`App` with EarlGrey 2.0. Note that you must pin the transitive dependency
+`eDistantObject` at 0.9.0.
```
tirodkar: Here is an extremely simple Podfile demonstrating integrating an app called
`App` with EarlGrey 2.0.

**Note:**  that you must pin the transitive dependency `eDistantObject` at
0.9.0.

tirodkar:
Thanks @adamhowardprice

PR #1307, opened 2020-05-20
System version error

willsaid-google:
The #else for #if defined(__IPHONE_11_0) should include what min system version is needed for backgrounding.


TestLib/EarlGreyImpl/EarlGreyImpl+XCUIApplication.m
```diff
@@ -54,7 +54,7 @@ - (BOOL)backgroundApplication {
   return [condition waitWithTimeout:10.0 pollInterval:kPollInterval];
 #else
   NSString *errorDescription =
-      @"Cannot perform backgrounding because it is not supported with the current system version.";
+      @"Cannot perform backgrounding because it is not supported with the current system version.\nUse at least iOS 11.";
```
tirodkar, line 57: @"Cannot perform backgrounding because it is not supported with the current system version."
@"Use at least iOS 11.";

PR #1316, opened 2020-05-28
Glossary Removal

willsaid-google:
Description Glossary Removal from Earl Grey Error Handling:

Recently, glossary contents have not been used at all, although they sometimes contain needed information (such as, what was the picker and the column that failed?)

If the glossary contents were important to the user, i added them to the exception reason.
If they weren't, i removed the glossary contents.

I also removed the glossary properties on GREYError, and the user info dictionary property as well being passed into I_GREYErrorMake, and the GREYPopulateErrorNoted macro, all of which were complicating error handling and making errors harder to read.

Example 1, where important info was not being shown, before:

picker-glossary-before

after:

picker-glossary-after

Example 2 
this time when the glossary contained too much info, before:

glossary-before

and after:

glossary-after

Of course, in this second example, a followup edit should remove the unnecessary Element Matcher key for this specific error code but that is outside the scope of the glossary removal.

Example 3 - Idling Resources
In this example, the output hasn't changed at all, although the glossary was removed.
idling

Example 4 - Stepper
If the stepper buttons aren't found, before:
```
Failed to find stepper buttons
in [Stepper]
```
Notice how it wasn't printing out the Stepper that is included in the glossary.
After:
```
Failed to find stepper buttons in stepper: 
<UIStepper: 0x7fdaaf52e030; frame = (205 35; 94 32); clipsToBounds = YES; opaque = NO; autoresize = RM+BM; layer = <CALayer: 0x6000026fa6a0>>
```

### DIFF
--- a/AppFramework/Action/GREYActions.m
+++ b/AppFramework/Action/GREYActions.m
@@ -573,24 +573,23 @@ static Protocol *gTextInputProtocol;
             }
             // Wait for keyboard to show up and any other UI changes to take effect.
             if (![GREYKeyboard waitForKeyboardToAppear]) {
-              NSString *description =
-                  @"Keyboard did not appear after tapping on [Element]. "
-                  @"Are you sure that tapping on this element will bring up the keyboard?";
               __block NSString *elementDescription;
               grey_dispatch_sync_on_main_thread(^{
                 elementDescription = [element grey_description];
               });
-              NSDictionary<NSString *, NSString *> *glossary = @{@"Element" : elementDescription};
-              I_GREYPopulateErrorNoted(errorOrNil, kGREYInteractionErrorDomain,
-                                       kGREYInteractionActionFailedErrorCode, description,
-                                       glossary);
+              NSString *description = [NSString stringWithFormat:
+                                       @"Keyboard did not appear after tapping on an element."
+                                       @"Are you sure that tapping on this element will bring up the keyboard?"
+                                       @"\nElement: %@", element];
+              I_GREYPopulateError(errorOrNil, kGREYInteractionErrorDomain,
+                                  kGREYInteractionActionFailedErrorCode, description);
               return NO;
             }
           }
 
           // If a position is given, move the text cursor to that position.
           __block id firstResponder = nil;
-          __block NSDictionary<NSString *, NSString *> *errorGlossary = nil;
+          __block NSMutableString *description = nil;
           grey_dispatch_sync_on_main_thread(^{
             firstResponder = [[expectedFirstResponderView window] firstResponder];
             if (position) {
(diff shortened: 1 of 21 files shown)


AppFramework/Action/GREYActions.m
```diff
@@ -573,24 +573,23 @@ + (void)initialize {
...
                 elementDescription = [element grey_description];
               });
-              NSDictionary<NSString *, NSString *> *glossary = @{@"Element" : elementDescription};
-              I_GREYPopulateErrorNoted(errorOrNil, kGREYInteractionErrorDomain,
-                                       kGREYInteractionActionFailedErrorCode, description,
-                                       glossary);
+              NSString *description = [NSString stringWithFormat:
+                                       @"Keyboard did not appear after tapping on an element."
+                                       @"Are you sure that tapping on this element will bring up the keyboard?"
+                                       @"\nElement: %@", element];
```
tirodkar, line 583:
```
NSString *description = 
    [NSString stringWithFormat:@"Keyboard did not appear after tapping on an element."
                               @"Are you sure that tapping on this element will bring up the keyboard?"
                               @"\nElement: %@", element];
```

AppFramework/Action/GREYActions.m
```diff
@@ -573,24 +573,23 @@ + (void)initialize {
...
+              I_GREYPopulateError(errorOrNil, kGREYInteractionErrorDomain,
+                                  kGREYInteractionActionFailedErrorCode, description);
               return NO;
             }
           }
 
           // If a position is given, move the text cursor to that position.
           __block id firstResponder = nil;
-          __block NSDictionary<NSString *, NSString *> *errorGlossary = nil;
+          __block NSMutableString *description = nil;
```
tirodkar, line 592: __block NSMutableString<NSString *, NSString *>

AppFramework/Action/GREYChangeStepperAction.m
```diff
@@ -190,12 +190,10 @@ - (GREYStepperButtons *)grey_stepperButtonsForStepper:(UIStepper *)stepper
...
-    NSString *description = [NSString stringWithFormat:@"Failed to find stepper buttons "
-                                                       @"in [Stepper]"];
-    NSDictionary<NSString *, NSString *> *glossary = @{@"[Stepper]" : [stepper description]};
-
-    I_GREYPopulateErrorNoted(error, kGREYInteractionErrorDomain,
-                             kGREYInteractionActionFailedErrorCode, description, glossary);
+    NSString *description = [NSString stringWithFormat:@"Failed to find stepper buttons in stepper:"
+                                                       @"%@", [stepper description]];
+    I_GREYPopulateError(error, kGREYInteractionErrorDomain,
+                             kGREYInteractionActionFailedErrorCode, description);
```
tirodkar, line 196: format

AppFramework/Action/GREYActions.m
chan-park: Instead of newline for each items of the glossary, how about adding numbering, or bullet points? I feel like that would be easier and cleaner than adding spaces, as well as parsing (if we have to). What do you think?

AppFramework/Assertion/GREYAssertions.m
```diff
@@ -37,22 +37,15 @@ @implementation GREYAssertions
...
-                                 kGREYInteractionElementNotFoundErrorCode, reason, glossary);
+        [reason appendFormat:@"Assertion with Matcher failed: No UI element was matched.\n"];
+        [reason appendFormat:@"Matcher: \n%@", [matcher description]];
+        I_GREYPopulateError(errorOrNil, kGREYInteractionErrorDomain,
+                            kGREYInteractionElementNotFoundErrorCode, reason);
       } else {
-        [reason appendFormat:@"Assertion with [Matcher] failed: UI [Element] failed to match "
-                             @"the following matcher(s): [Mismatch]"];
-        glossary[@"Matcher"] = [matcher description];
-        glossary[@"Element"] = [element grey_description];
```
willsaid-google, line 51: Add these back in to the reason

PR #1319, opened 2020-05-28
Fix method call in demo project since deeplink function name changed

brettfazio:
The name for the `openDeeplinkURL(url, in: application)` function changed and it wasn't updated in the demo project. I just updated the line in the demo project.

### DIFF
--- a/TestLib/Swift/EarlGrey.swift
+++ b/TestLib/Swift/EarlGrey.swift
@@ -153,7 +153,7 @@ public struct EarlGrey {
     line: UInt = #line
   ) throws {
     try EarlGreyImpl.invoked(fromFile: file.description, lineNumber: line)
-      .openDeepLinkURL(url, with: application)
+      .openDeeplinkURL(url, in: application)
   }
 
   public static func remoteClassInApp(


brettfazio:
@googlebot I signed it!

tirodkar:
Thanks for this Brett! I'll test this internally and reply back.

tirodkar:
You've updated the TestLib file - I believe the change was in the Demo project?

brettfazio:
The change was in the Demo project, but the demo project links the `EarlGrey.swift` file directly into it. If you look at the commit that adds that line into EarlGrey.swift (https://github.com/google/EarlGrey/commit/44a82f9d4f2f2b8987988402bffff29d73de102c#diff-986ac37ff4eb69a0b14ac632de6e0a4bR156) and compare it to the method signature in EarlGrey.h:

`- (BOOL)openDeeplinkURL:(NSString *)URL
          inApplication:(XCUIApplication *)application
                  error:(NSError **)error;`

It doesn't match. If you try doing pod install on the demo project and running it you should see the issue in EarlGrey.swift pop up.

tirodkar:
Got it. We'd need to do a new release with EarlGrey.swift and the new changes for OSS users. Can you import the EarlGrey.swift file from the Demo till then?

PR #1320, opened 2020-05-29
fixed 4 error typos

willsaid-google:


Tests/Functional/Sources/ErrorHandlingTest.m
```diff
@@ -147,8 +147,7 @@ - (void)testRotationDescriptionGlossary {
 - (void)testKeyboardDismissalError {
   NSError *error;
   [EarlGrey dismissKeyboardWithError:&error];
-  NSString *keyboardErrorString = @"Failed to dismiss keyboard since it was not showing. Internal "
-                                  @"Error: Failed to dismiss keyboard as it was not shown.";
```
tirodkar, line 151: Is this no longer showing?

willsaid-google: It's redundant, right? Why do we need to say the same error twice?

PR #1330, opened 2020-06-02
Error Formatting: ElementNotFoundError

willsaid-google:
This change tracks the formatting of the kGREYInteractionElementNotFoundErrorCode, for both search actions and non-search actions, to conform to the console output shown below.

Chart of how error formatting works for Element Not Found, before and after:
Copy of Error Handling

Note that the avg execution time before & after for each of the 3 following errors did not change by a statistically significant amount.

### **Element Not Found assertion failure from a Search Action** 
**Console Output, Before:**
```
Exception Name: com.google.earlgrey.ElementInteractionErrorDomain
Exception Reason: Search action failed: Interaction cannot continue because the desired element was not found.

Element Matcher:  ...
Exception with Assertion: {
"Assertion Criteria" : "assertWithMatcher:sufficientlyVisible(Expected: 0.750000, Actual: 0.000000)",

"Element Matcher" : "(((respondsToSelector(isAccessibilityElement) && isAccessibilityElement) && accessibilityLabel('Label 2')) && interactable Point:{nan, nan} && sufficientlyVisible(Expected: 0.750000, Actual: 0.000000))",

"Recovery Suggestion" : "Check if the element exists in the UI hierarchy printed below. If it exists, adjust the matcher so that it accurately matches element.",

"Search API Info" : "Search action: <GREYScrollAction: 0x600003457fc0>. 
Search action element matcher: ((respondsToSelector(isAccessibilityElement) && isAccessibilityElement) && accessibilityLabel('Invalid Scroll View'))."
}

Bundle ID: com.google.com.Functional.xctrunner

Stack Trace: ...

Screenshots: {
  "App-side Screenshot at Point-of-Failure" : "a.png",
  "Test-side Screenshot at Failure" : "b.png"
}

UI Hierarchy: ...
```

### **Element Not Found assertion failure from a Search Action** 
**Console Output, After:**
```
EarlGrey Encountered an Error:

Search action failed: Interaction cannot continue because the desired element was not found. 

Check if the element exists in the UI hierarchy printed below. If it exists, adjust the matcher so that it accurately matches the element.

Element Matcher:
(((respondsToSelector(isAccessibilityElement) && isAccessibilityElement) && accessibilityLabel('Label 2')) && interactable Point:{nan, nan} && sufficientlyVisible(Expected: 0.750000, Actual: 0.000000))

Assertion Criteria: assertWithMatcher:sufficientlyVisible(Expected: 0.750000, Actual: 0.000000)

Search API Info
Search action: <GREYScrollAction: 0x600003457fc0>. 
Search action element matcher: ((respondsToSelector(isAccessibilityElement) && isAccessibilityElement) && accessibilityLabel('Invalid Scroll View')).

App-side Screenshot at Point-of-Failure: a.png

Test-side Screenshot at Failure : b.png

=====UI Hierarchy===== 

=====Stack Trace===== 
```

### **Element Not Found assertion failure NOT from a Search Action** 
**Console Output, Before:**
```
Exception Name: com.google.earlgrey.ElementInteractionErrorDomain

Exception Reason: Interaction cannot continue because the desired element was not found.
Element Matcher: Cannot find UI Element.
Exception with Assertion: {
"Assertion Criteria" : "assertWithMatcher:isNotNil",

"Element Matcher" : "((kindOfClass('UILabel') || kindOfClass('UITextField') || kindOfClass('UITextView')) && hasText('Basic Views'))",

"Recovery Suggestion" : "Check if the element exists in the UI hierarchy printed below. If it exists, adjust the matcher so that it accurately matches element."
}

Bundle ID: com.google.com.Functional.xctrunner

Stack Trace: ...

Screenshots: {
  "App-side Screenshot at Point-of-Failure" : "a.png",
  "Test-side Screenshot at Failure" : "b.png"
}

UI Hierarchy: ...
```

### **Element Not Found assertion failure NOT from a Search Action** 
**Console Output, After:**
```
EarlGrey Encountered an Error:

Interaction cannot continue because the desired element was not found. 

Check if the element exists in the UI hierarchy printed below. If it exists, adjust the matcher so that it accurately matches the element.

Element Matcher:
((kindOfClass('UILabel') || kindOfClass('UITextField') || kindOfClass('UITextView')) && hasText('Basic Views'))

Assertion Criteria: assertWithMatcher:isNotNil

App-side Screenshot at Point-of-Failure: a.png

Test-side Screenshot at Failure : b.png

=====UI Hierarchy===== 

=====Stack Trace===== 
```

### **Element Not Found _action_ failure** 
**Console Output, Before:**
```
Exception Name: com.google.earlgrey.ElementInteractionErrorDomain

Exception Reason: Interaction cannot continue because the desired element was not found.
Element Matcher: Cannot find UI Element.
Exception with Action: {
    "Action Name" : "Execute JavaScript",
    "Element Matcher" : "(respondsToSelector(accessibilityIdentifier) && accessibilityID('TestWKWebView'))",
      "Recovery Suggestion" : "Check if the element exists in the UI hierarchy printed below. If it exists, adjust the matcher so that it accurately matches element."
}

Bundle ID: com.google.com.Functional.xctrunner

Stack Trace: ...

Screenshots: {
  "App-side Screenshot at Point-of-Failure" : "a.png",
  "Test-side Screenshot at Failure" : "b.png"
}

UI Hierarchy: ...
```

### **Element Not Found _action_ failure** 
**Console Output, After:**
```
EarlGrey Encountered an Error:

Interaction cannot continue because the desired element was not found. 

Check if the element exists in the UI hierarchy printed below. If it exists, adjust the matcher so that it accurately matches the element.

Element Matcher:
(respondsToSelector(accessibilityIdentifier) && accessibilityID('TestWKWebView'))

Action Name: Execute JavaScript

App-side Screenshot at Point-of-Failure: a.png

Test-side Screenshot at Failure : b.png

=====UI Hierarchy===== 

=====Stack Trace===== 
```

Note that "EarlGrey Encountered an Error:" is temporary, until all error codes are using GREYErrorFormatter.


GREYErrorFormatter.m
```diff
@@ -0,0 +1,120 @@
...
+
+@property(readonly, nonatomic) GREYError *error;
+
+@end
+
+@implementation GREYErrorFormatter
+
+#pragma mark - Class Methods
+
++ (NSString *)FormattedHierarchy:(nonnull NSString *)hierarchy {
```
chan-park, line 25: should be lowercase

GREYErrorFormatter.m
```diff
@@ -0,0 +1,120 @@
...
+    NSMutableArray *logger = [[NSMutableArray alloc] init];
```
chan-park, line 26: Specify generics

GREYErrorFormatter.m
```diff
@@ -0,0 +1,120 @@
...
+    
+    [logger addObject:@"UI Hierarchy (ordered by window level, back to front):\n"];
+
+    NSString *windowLegend = @"[Window 1]";
+    NSString *axLegend = @"[AX]";
+    NSString *uieLegend = @"[UIE]";
+
+    NSDictionary *legendLabels = @{
```
chan-park, line 34: Moving forwards, when you're initializing containers, you should always specify generics.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,120 @@
...
+  if (self) {
+    _error = error;
+  }
+  return self;
+}
+
+#pragma mark - Public Methods
+
+- (NSString *)formattedDescription {
+  if (_error.domain == kGREYInteractionErrorDomain &&
```
chan-park, line 63: In ObjC, you have to use `-[NSString isEqualToString:]` when comparing strings. == will only compare the addresses.

TestLib/EarlGreyImpl/GREYElementInteractionErrorHandler.m
```diff
@@ -78,7 +78,12 @@ void GREYHandleInteractionError(__strong GREYError *interactionError,
       id<GREYFailureHandler> failureHandler =
           [NSThread mainThread].threadDictionary[GREYFailureHandlerKey];
       // TODO(b/147072566): Will show up a (null) in rotation.
-      [failureHandler handleException:exception details:matcherDetails];
+      if ([reason containsString:@"the desired element was not found"]) {
+        /// Eventually this check will not be needed, and will be run for every exception.
```
chan-park, line 82: Not sure what you mean by this comment.

GREYErrorFormatter.h
```diff
@@ -0,0 +1,44 @@
...
+ * For a given @c error, creates a GREYErrorFormatter to provide a formattedDescription.
+ *
+ * @param error the GREYError used to format the description
+ *
+ * @return An instance of GREYErrorFormatter
+*/
+- (instancetype)initWithError:(GREYError *)error;
+
+/**
+ * Creates the human-readable formatted string for the GREYError,
```
chan-park, line 34: Creates a human-readable*

GREYErrorFormatter.h
```diff
@@ -0,0 +1,44 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface GREYErrorFormatter: NSObject
+
+/**
+ * For a given @c hierarchy, creates the human-readable hierarchy
+ * including the legend.
+ *
+ * @param hierarchy the string representation of the App UI Hierarchy
```
chan-park, line 18: The

GREYErrorFormatter.h
```diff
@@ -0,0 +1,44 @@
...
+//
+
+#import "GREYError.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface GREYErrorFormatter: NSObject
+
+/**
+ * For a given @c hierarchy, creates the human-readable hierarchy
```
chan-park, line 15: For a given @c hierarchy, create a human-readable hierarchy...

GREYErrorFormatter.h
```diff
@@ -0,0 +1,44 @@
...
+ * including the legend.
+ *
+ * @param hierarchy the string representation of the App UI Hierarchy
+ *
+ * @return The formatted hierarchy and legend to be printed on the console
+*/
++ (NSString *)FormattedHierarchy:(nonnull NSString *)hierarchy;
+
+/**
+ * For a given @c error, creates a GREYErrorFormatter to provide a formattedDescription.
```
chan-park, line 25: Same as above. @c formattedDescription.

willsaid-google:
@chan-park Thanks for the review - i fixed all the edits.

AppFramework/Core/GREYElementInteraction.m
```diff
@@ -806,7 +806,13 @@ - (NSError *)grey_errorToReturnForInteractionError:(GREYError *)interactionError
   NSString *hierarchy = [self grey_unifyAndExtractHierarchyFromError:interactionError];
 
   // Add information such as element matcher and any nested error info.
+  // Copy over the matcher details from the error info dictionary.
   NSMutableDictionary<NSString *, id> *userInfo = [[NSMutableDictionary alloc] init];
+  if ([interactionError isKindOfClass:[GREYError class]]) {
```
tirodkar, line 811: Can you explain more why this is needed?

CommonLib/Error/GREYError.m
```diff
@@ -108,6 +110,11 @@ + (instancetype)errorWithDomain:(NSString *)domain
 }
 
 - (NSString *)description {
+  if (self.domain == kGREYInteractionErrorDomain &&
```
tirodkar, line 113: Move to a static function in GREYErrorFormatter. You could even just have this as checked in formattedDescription.

NSString *des = [[[GREYErrorFormatter alloc] initWithError:self] formattedDescription];
if (!des) {
  return [GREYObjectFormatter formatDictionary:[self grey_descriptionDictionary]
                                          indent:kGREYObjectFormatIndent
                                       hideEmpty:YES...
}

GREYErrorFormatter.h
```diff
@@ -0,0 +1,44 @@
+//
```
tirodkar, line 1: Change this to the License we have in other files

GREYErrorFormatter.m
```diff
@@ -0,0 +1,120 @@
+//
+//  GREYErrorFormatter.m
+//  CommonLib
+//
+//  Created by Will Said on 6/1/20.
```
tirodkar, line 5: Same.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,120 @@
...
+    _error = error;
+  }
+  return self;
+}
+
+#pragma mark - Public Methods
+
+- (NSString *)formattedDescription {
+  if ([_error.domain isEqualToString:kGREYInteractionErrorDomain] &&
+      _error.code == kGREYInteractionElementNotFoundErrorCode) {
```
tirodkar, line 64: This is replicated in GREYError, you can just keep this here and return nil instead of crashing.

TestLib/EarlGreyImpl/GREYElementInteractionErrorHandler.m
```diff
@@ -78,7 +78,12 @@ void GREYHandleInteractionError(__strong GREYError *interactionError,
       id<GREYFailureHandler> failureHandler =
           [NSThread mainThread].threadDictionary[GREYFailureHandlerKey];
       // TODO(b/147072566): Will show up a (null) in rotation.
-      [failureHandler handleException:exception details:matcherDetails];
+      if ([reason containsString:@"the desired element was not found"]) {
```
tirodkar, line 81: interactionError should be the raw error returned from the app side - is it not giving the ElementNotFoundErrorCode here?

TestLib/Exception/GREYDefaultFailureHandler.m
```diff
@@ -89,17 +90,34 @@ - (void)handleException:(GREYFrameworkException *)exception details:(NSString *)
...
-                                                     failureName:exception.name
-                                                        filePath:_fileName
-                                                      lineNumber:_lineNumber
-                                                    functionName:nil
-                                                      stackTrace:stackTrace
-                                                  appScreenshots:screenshotPaths
-                                                       hierarchy:appUIHierarchy
-                                                          format:@"%@", logMessage];
-  [currentTestCase grey_markAsFailedAtLine:_lineNumber inFile:_fileName description:log];
+  if ([exception.reason containsString:@"the desired element was not found"]) {
```
tirodkar, line 93: How about moving this code to a new method?

AppFramework/Core/GREYElementInteraction.m
willsaid-google: In case a user or a test case passes in an NSError ** into an EarlGrey method. Without this check, the error handling tests were crashing.

willsaid-google:
Fixed/Responded to all of Aditya's comments @tirodkar

TestLib/Exception/GREYDefaultFailureHandler.m
```diff
@@ -81,24 +93,57 @@ - (void)handleException:(GREYFrameworkException *)exception details:(NSString *)
...
-                                                    functionName:nil
-                                                      stackTrace:stackTrace
-                                                  appScreenshots:screenshotPaths
-                                                       hierarchy:appUIHierarchy
-                                                          format:@"%@", logMessage];
+- (void)handleException:(GREYFrameworkException *)exception details:(NSString *)details {
+  GREYThrowOnNilParameter(exception);
+  id currentTestCase = [XCTestCase grey_currentTestCase];
+  GREYFailureScreenshots *screenshotPaths = [self screenshotPathsForException:exception];
+  NSArray *stackTrace = [NSThread callStackSymbols];
```
chan-park, line 139: Please add generics for all container types across all affected code.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,200 @@
...
+  if ([error isKindOfClass:[GREYError class]]) {
+    NSMutableDictionary *mutableDescriptions = [descriptions mutableCopy];
+    [mutableDescriptions removeObjectForKey:kErrorUserInfoKey];
+    [mutableDescriptions removeObjectForKey:kErrorErrorInfoKey];
+    [mutableDescriptions removeObjectForKey:kErrorStackTraceKey];
+    [mutableDescriptions removeObjectForKey:kErrorAppUIHierarchyKey];
+    [mutableDescriptions removeObjectForKey:kErrorAppScreenShotsKey];
+    descriptions = mutableDescriptions;
+  }
+  [errorStack addObject:descriptions];
```
chan-park, line 108: Do you want to change this back to immutable or not?

GREYErrorFormatter.m
```diff
@@ -0,0 +1,200 @@
...
+#pragma mark - Private Methods
+
+- (NSString *)elementNotFoundDescription {
+  NSMutableArray<NSString *> *logger = [[NSMutableArray alloc] init];
+  
+  // exception reason
+  [logger addObject:[NSString stringWithFormat:@"\n%@\n", _error.localizedDescription]];
+
+  // recovery suggestion
+  if (_error.userInfo[kErrorDetailRecoverySuggestionKey]) {
```
chan-park, line 138: Store the value to a local variable and reuse that instead of fetching it twice.

GREYErrorFormatter.h
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * For a given @c hierarchy, create a human-readable hierarchy
+ * including the legend.
+ *
+ * @param hierarchy The string representation of the App UI Hierarchy
+ *
+ * @return The formatted hierarchy and legend to be printed on the console
+*/
++ (NSString *)formattedHierarchy:(nonnull NSString *)hierarchy;
```
AlbertWang0116, line 31: Prefer C API to class method. Class method created redundant symbols/encoding/runtime, so let's avoid it if it's not necessary. Extended reading:
https://medium.com/@dmaclach/objective-c-encoding-and-you-866624cc02de

AppFramework/Core/GREYElementInteraction.m
```diff
@@ -193,7 +194,7 @@ - (void)matchElementsWithTimeout:(CFTimeInterval)timeout
   } else if (searchActionError) {
     NSString *searchActionDescription = searchActionError.localizedDescription;
     NSString *description =
-        [NSString stringWithFormat:@"Search action failed: %@\n", searchActionDescription];
```
tirodkar, line 196: Was this an accidental change?

AppFramework/Core/GREYElementInteraction.m
tirodkar: Could you share the details of the failing test? This means that we have an NSError being passed into GREYElementInteraction which should not happen as the TestLib GREYElementInteractionProxy creates a GREYError to be sent.

CommonLib/Error/GREYError.m
```diff
@@ -16,6 +16,7 @@
 
 #import "GREYError.h"
 
+#import "GREYErrorFormatter.h"
 #import "GREYObjectFormatter.h"
```
tirodkar, line 20: Is this used anywhere?

GREYErrorFormatter.h
```diff
@@ -0,0 +1,66 @@
+//
+// Copyright 2017 Google Inc.
```
tirodkar, line 2: 2020

GREYErrorFormatter.h
```diff
@@ -0,0 +1,66 @@
...
+// limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+
+@class GREYError;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface GREYErrorFormatter: NSObject
```
tirodkar, line 23: Add description for the class.

GREYErrorFormatter.h
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * For a given @c error, create a GREYErrorFormatter to provide a formattedDescription.
```
tirodkar, line 26: Creates a GREYErrorFormatter for formatting the description of a GREYError.

GREYErrorFormatter.h
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param error The GREYError used to format the description
+ *
+ * @return An instance of GREYErrorFormatter
+ */
+- (instancetype)initWithError:(GREYError *)error;
```
tirodkar, line 32: Add NS_DESIGNATED_INITIALIZER and NS_UNAVAILABLE for init.

GREYErrorFormatter.h
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Create a human-readable formatted string for the GREYError,
```
tirodkar, line 35: You can club this with the @return 

/**
 * @return A custom formatting of the GREYError tailored to the type of error.
*/

GREYErrorFormatter.h
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * For a given @c hierarchy, create a human-readable hierarchy
+ * including the legend.
+ *
+ * @param hierarchy The string representation of the App UI Hierarchy
+ *
+ * @return The formatted hierarchy and legend to be printed on the console
+ */
+//+ (NSString *)formattedHierarchy:(nonnull NSString *)hierarchy;
```
tirodkar, line 51: rm. What is the FormatterHierarchy for?

GREYErrorFormatter.h
```diff
@@ -0,0 +1,66 @@
...
+NSString *GREYFormattedHierarchy(NSString *hierarchy);
+
+/**
+ * Determine whether this error's code and domain are supported
+ * by the new GREYErrorFormatter formattedDescription.
+ * If NO, then default to using the existing dictionary object formatting.
+ *
+ * @return YES if the new formatting should be used for this error
+ */
+//+ (BOOL)shouldUseErrorFormatterForError:(GREYError *)error;
```
tirodkar, line 61: Remove commented code.

GREYErrorFormatter.h
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @return The formatted hierarchy and legend to be printed on the console
+ */
+//+ (NSString *)formattedHierarchy:(nonnull NSString *)hierarchy;
+NSString *GREYFormattedHierarchy(NSString *hierarchy);
+
+/**
+ * Determine whether this error's code and domain are supported
+ * by the new GREYErrorFormatter formattedDescription.
+ * If NO, then default to using the existing dictionary object formatting.
```
tirodkar, line 57: @c NO, @c YES

GREYErrorFormatter.h
```diff
@@ -0,0 +1,66 @@
...
+ * @param hierarchy The string representation of the App UI Hierarchy
+ *
+ * @return The formatted hierarchy and legend to be printed on the console
+ */
+//+ (NSString *)formattedHierarchy:(nonnull NSString *)hierarchy;
+NSString *GREYFormattedHierarchy(NSString *hierarchy);
+
+/**
+ * Determine whether this error's code and domain are supported
+ * by the new GREYErrorFormatter formattedDescription.
```
tirodkar, line 56: GREYErrorFormatter::formattedDescription

GREYErrorFormatter.m
```diff
@@ -0,0 +1,136 @@
+//
+// Copyright 2017 Google Inc.
```
tirodkar, line 2: 2020*

GREYErrorFormatter.m
```diff
@@ -0,0 +1,136 @@
...
+#import "GREYErrorFormatter.h"
+
+#import "GREYObjectFormatter.h"
+#import "GREYInteraction.h"
+#import "GREYErrorConstants.h"
+#import "NSError+GREYCommon.h"
+#import "GREYError+Private.h"
+
+@interface GREYErrorFormatter ()
+
```
tirodkar, line 26: rm empty lines. Comment the @interface.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,136 @@
...
+#import "GREYErrorFormatter.h"
+
+#import "GREYObjectFormatter.h"
+#import "GREYInteraction.h"
+#import "GREYErrorConstants.h"
+#import "NSError+GREYCommon.h"
+#import "GREYError+Private.h"
+
+@interface GREYErrorFormatter ()
```
tirodkar, line 25: rm whitespace

GREYErrorFormatter.m
```diff
@@ -0,0 +1,136 @@
...
+  }
+  return [GREYObjectFormatter formatDictionary:[_error grey_descriptionDictionary]
+                                        indent:kGREYObjectFormatIndent
+                                     hideEmpty:YES
+                                      keyOrder:nil];
+}
+
+#pragma mark - Private Methods
+
+- (NSString *)elementNotFoundDescription {
```
tirodkar, line 57: This method name is too tightly coupled to the error being described. For new error codes, this would mean we would have to have a function for each error code. 

I think a better way would be if we print any of the below information if it is provided to us and leave it to the error creator to actually fill out the error. WDYT?

GREYErrorFormatter.m
```diff
@@ -0,0 +1,136 @@
...
+#pragma mark - Static Functions
+
+NSString *GREYFormattedHierarchy(NSString * hierarchy) {
+  if (!hierarchy) {
+    return nil;
+  }
+  NSMutableArray<NSString*> *logger = [[NSMutableArray alloc] init];
+  
+  [logger addObject:@"UI Hierarchy (ordered by window level, back to front):\n"];
+
```
tirodkar, line 110: remove these newlines to make code compact.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,136 @@
...
+  }
+  NSMutableArray<NSString*> *logger = [[NSMutableArray alloc] init];
+  
+  [logger addObject:@"UI Hierarchy (ordered by window level, back to front):\n"];
+  
+  NSString *windowLegend = @"[Window 1]";
+  NSString *axLegend = @"[AX]";
+  NSString *uieLegend = @"[UIE]";
+  
+  NSDictionary<NSString *, NSString *> *legendLabels = @{
```
tirodkar, line 115: Can we just switch this into a static string?

AppFramework/Core/GREYElementInteraction.m
willsaid-google: Several tests would fail without this check. For example, testCustomNSErrorInAction in ErrorHandlingTest

AppFramework/Core/GREYElementInteraction.m
willsaid-google: No, this is intentional. This newline character was interfering with the way ErrorFormatter inserts a new line character between each key.

CommonLib/Error/GREYError.m
willsaid-google: line 112

GREYErrorFormatter.h
willsaid-google: Good catch!

GREYErrorFormatter.m
willsaid-google: I'm not sure what comment to leave. Shouldn't the comment be in the .h interface instead?

GREYErrorFormatter.m
willsaid-google: Updated. I first look up the set of keys that should be shown for a given code/domain, and then pass that to another method that prints out the values of those keys.

Tests/Functional/Sources/ErrorHandlingTest.m
```diff
@@ -125,10 +126,10 @@ - (void)testDescriptionForSearchActionWithAssertion {
       onElementWithMatcher:grey_accessibilityLabel(@"Invalid Scroll View")]
       assertWithMatcher:grey_sufficientlyVisible()
                   error:&error];
-  NSString *searchAPIDescription = @"\"Search API Info\" : \"Search Action:";
```
tirodkar, line 128: Can you show me how this is printed?

TestLib/Exception/GREYFailureFormatter.h
```diff
@@ -76,7 +50,7 @@ NS_ASSUME_NONNULL_BEGIN
  * @param stackTrace     The stack trace of the executable when the error happens.
  * @param appScreenshots The dictionary of app screenshot names and image paths
  *                       when the error happens.
- * @param format         Extra message to be included in the message.
+ * @param errorDescription         Extra message to be included in the message.
```
tirodkar, line 53: Fix formatting.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,195 @@
...
+                                     hideEmpty:YES
+                                      keyOrder:nil];
+}
+
+#pragma mark - Private Methods
+
+- (NSSet<NSString *> *)loggerKeys {
+  if ([_error.domain isEqualToString:kGREYInteractionErrorDomain] &&
+       _error.code == kGREYInteractionElementNotFoundErrorCode) {
+    return [[NSSet alloc] initWithArray:@[kErrorFormatterExceptionReasonKey,
```
tirodkar, line 88: Can you use NSSet setWithObjects?

GREYErrorFormatter.m
```diff
@@ -0,0 +1,195 @@
...
+  }
+  return [GREYObjectFormatter formatDictionary:[_error grey_descriptionDictionary]
+                                        indent:kGREYObjectFormatIndent
+                                     hideEmpty:YES
+                                      keyOrder:nil];
+}
+
+#pragma mark - Private Methods
+
+- (NSSet<NSString *> *)loggerKeys {
```
chan-park, line 85: Can we cache and reuse the same set? Instead of creating a new one every time.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,195 @@
...
+    NSArray<NSString *> *stackTrace = _error.stackTrace;
+    if (stackTrace) {
+      [logger addObject:[NSString stringWithFormat:@"%@: %@\n", kErrorStackTraceKey, stackTrace]];
+    }
+  }
+  
+  return [logger componentsJoinedByString:@"\n"];
+}
+
+#pragma mark - Static Functions
```
chan-park, line 163: Static function in C means it's scoped to this file. This function is public right?

GREYErrorFormatter.m
```diff
@@ -0,0 +1,195 @@
...
+                                                            hideEmpty:NO
+                                                             keyOrder:keyOrder];
+  [logger addObject:[NSString stringWithFormat:@"%@: %@\n", kHierarchyLegendKey,
+                     legendDescription]];
+  [logger addObject:hierarchy];
+  return [logger componentsJoinedByString:@"\n"];
+}
+
+BOOL GREYShouldUseErrorFormatterForError(GREYError *error) {
+  return [error.domain isEqualToString:kGREYInteractionErrorDomain] &&
```
chan-park, line 191: This might be a nitpicking but checking error.code first then error.domain is more efficient.

TestLib/Exception/GREYDefaultFailureHandler.m
```diff
@@ -81,24 +93,57 @@ - (void)handleException:(GREYFrameworkException *)exception details:(NSString *)
...
-  if (!appUIHierarchy) {
-    appUIHierarchy = [GREYElementHierarchy hierarchyString];
+- (NSString *)consoleLogForException:(GREYFrameworkException *)exception
+                             details:(NSString *)details
+                     screenshotPaths:(GREYFailureScreenshots *)screenshotPaths
+                      appUIhierarchy:(NSString *)appUIHierarchy
+                     currentTestCase:(id)currentTestCase
+                          stackTrace:(NSArray *)stackTrace {
+  if ([exception.reason containsString:@"the desired element was not found"]) {
+    /// As GREYErrorFormatter adds support for more error codes,
```
chan-park, line 106: //*

TestLib/Exception/GREYDefaultFailureHandler.m
```diff
@@ -81,24 +93,57 @@ - (void)handleException:(GREYFrameworkException *)exception details:(NSString *)
...
+    NSMutableString *formattedOutput = [details mutableCopy];
+    /// Appends screenshots and ui hierarchy to the console output if they didn't already exist.
+    if (!exception.userInfo[kErrorDetailAppScreenshotsKey]) {
+      for (NSString *key in screenshotPaths.allKeys) {
+        [formattedOutput appendString:[NSString stringWithFormat:@"\n%@: %@\n",
+                                       key, screenshotPaths[key]]];
+      }
+    }
+    if (!exception.userInfo[kErrorDetailAppUIHierarchyKey]) {
+    [formattedOutput appendString:[NSString stringWithFormat:@"\n%@\n",
```
chan-park, line 117: Use [`-[NSMutableString appendFormat:]`](https://developer.apple.com/documentation/foundation/nsmutablestring/1497308-appendformat?language=objc). please apply this to all other changes.

GREYErrorFormatter.h
```diff
@@ -0,0 +1,69 @@
...
+NSString *GREYFormattedHierarchy(NSString *hierarchy);
+
+/**
+ * Determines whether this error's code and domain are supported
+ * by the new GREYErrorFormatter::formattedDescription.
+ * If @c NO, then default to using the existing dictionary object formatting.
+ *
+ * @return @c YES if the new formatting based on the code and domain should be used for this error
+ */
+BOOL GREYShouldUseErrorFormatterForError(GREYError *error);
```
AlbertWang0116, line 65: Is this temporary?

GREYErrorFormatter.m
```diff
@@ -0,0 +1,195 @@
...
+ * All keys that can be added to the description logger in GREYErrorFormatter.
+ */
+static NSString *const kErrorFormatterExceptionReasonKey    = @"Exception Reason";
+static NSString *const kErrorFormatterRecoverySuggestionKey = @"Recovery Suggestion";
+static NSString *const kErrorFormatterElementMatcherKey     = @"Element Matcher";
+static NSString *const kErrorFormatterSearchActionInfoKey   = @"Search Action Info";
+static NSString *const kErrorFormatterScreenshotPathsKey    = @"Screenshot Paths";
+static NSString *const kErrorFormatterUnderlyingErrorKey    = @"Underlying Error";
+static NSString *const kErrorFormatterUIHierarchyKey        = @"UI Hierarchy";
+static NSString *const kErrorFormatterStackTraceKey         = @"Stack Trace";
```
AlbertWang0116, line 38: Can we use enum and bitwise computation instead?

GREYErrorFormatter.m
```diff
@@ -0,0 +1,195 @@
...
+  }
+  GREYFatalAssertWithMessage(false, @"Error Domain and Code Not Yet Supported");
+}
+
+- (NSString *)loggerDescriptionForKeys:(NSSet<NSString *> *)keys {
+  NSMutableArray<NSString *> *logger = [[NSMutableArray alloc] init];
+  
+  if ([keys containsObject:kErrorFormatterExceptionReasonKey]) {
+    NSString *exceptionReason = _error.localizedDescription;
+    [logger addObject:[NSString stringWithFormat:@"\n%@\n", exceptionReason]];
```
AlbertWang0116, line 105: A lot of format strings like "\n%@\n", "%@\n" are used. I'm wondering if the following helper method will be more readable:

[NSString grey_errorEntryWithSpacing:YES/NO format:@"%@", exceptionReason];

Or we can make it more specific:
[NSString grey_errorEntry:foo] ->`@"\n%@\n", foo`
[NSString grey_oneLinedErrorInfo:info forKey:key] -> `@"\n%@: %@\n", key, info`
[NSString grey_twoLinedErrorInfo:info forKey:key] -> `@"\n%@\n%@\n", key, info`

GREYErrorFormatter.m
```diff
@@ -0,0 +1,195 @@
...
+    for (NSString *key in _error.appScreenshots.allKeys) {
+      NSString *screenshotPath = _error.appScreenshots[key];
+      [logger addObject:[NSString stringWithFormat:@"%@: %@\n", key, screenshotPath]];
+    }
+  }
+  
+  if ([keys containsObject:kErrorFormatterUnderlyingErrorKey]) {
+    NSString *nestedError = _error.nestedError.description;
+    if (nestedError) {
+      [logger addObject:[NSString stringWithFormat:@"%@: \n%@\n", kErrorFormatterUnderlyingErrorKey,
```
AlbertWang0116, line 141: redundant space in "%@: \n%@\n"?

Tests/Functional/Sources/ErrorFormattingTest.m
```diff
@@ -0,0 +1,130 @@
...
+                               @"the element.\n"
+                               @"\n"
+                               @"Element Matcher:\n"
+                               @"((kindOfClass('UILabel') || kindOfClass('UITextField') || "
+                               @"kindOfClass('UITextView')) && hasText('Basic Views'))\n"
+                               @"\n"
+                               @"App-side Screenshot at Point-of-Failure:";
+  NSString *expectedDetails2 = @"UI Hierarchy (ordered by window level, back to front):";
+  NSString *expectedDetails3 = @"Stack Trace: (";
+  for (NSString *expectation in @[expectedDetails1, expectedDetails2, expectedDetails3]) {
```
AlbertWang0116, line 88: Expand this. In the case of failure, the current format doesn't give you useful information, just saying "[handler.details containsString:expectation] is false".

Same below.

Tests/Functional/Sources/ErrorHandlingTest.m
willsaid-google: We talked about this offline, but it contains whitespace padding of ~10 characters and should be removed, and I believe you are tracking that change already.

TestLib/Exception/GREYFailureFormatter.h
willsaid-google: It is not an issue in the actual code, just a github formatting issue. Checkout the branch and let me know if you still want it to be changed.

GREYErrorFormatter.m
willsaid-google: But won't it only be called once? Why would someone instantiate an ErrorFormatter and repeatedly call this method?

GREYErrorFormatter.m
willsaid-google: But it is always more likely to fail on the domain, since it's a string that's different for most errors, whereas the code is usually 0

GREYErrorFormatter.h
willsaid-google: Yes. I'll add a comment for that.

GREYErrorFormatter.m
willsaid-google: Would there be any additional benefit?

GREYErrorFormatter.m
willsaid-google: No, because this would say "Underlying Error: ..."

Tests/Functional/Sources/ErrorFormattingTest.m
willsaid-google: Good point.

GREYErrorFormatter.m
willsaid-google: You're right then - the string formatting is not very readable, so i made a new helper method for it

GREYErrorFormatter.m
willsaid-google: That is what i initially did, but then passing around the array of the enumeration values became a larger issue than the benefit it brought. Do you know of a workaround for that?

AppFramework/Core/GREYElementInteraction.m
```diff
@@ -802,13 +803,23 @@ - (BOOL)grey_handleFailureOfAssertion:(id<GREYAssertion>)assertion
 
 - (NSError *)grey_errorToReturnForInteractionError:(GREYError *)interactionError
                                         withReason:(NSString *)reason {
+  // TODO: @wsaid remove the `reason` parameter and replace it with the errorDetails dict
```
tirodkar, line 806: //TODO(wsaid): Remove the reason parameter and replace with the errorDetails dictionary.

AppFramework/Core/GREYElementInteraction.m
```diff
@@ -802,13 +803,23 @@ - (BOOL)grey_handleFailureOfAssertion:(id<GREYAssertion>)assertion
...
   // Add information such as element matcher and any nested error info.
   NSMutableDictionary<NSString *, id> *userInfo = [[NSMutableDictionary alloc] init];
   [userInfo setValue:interactionError.localizedDescription forKey:NSLocalizedDescriptionKey];
   [userInfo setValue:reason forKey:NSLocalizedFailureReasonErrorKey];
+  
+  // Copy over the matcher details from the error info dictionary if the error is a GREYError;
+  // else, passing in an NSError will crash here
+  if ([interactionError isKindOfClass:[GREYError class]]) {
+    for (NSString *key in interactionError.errorInfo) {
+      userInfo[key] = interactionError.errorInfo[key];
```
tirodkar, line 819: Can you not use [NSMutableDictionary addEntriesFromDictionary:] ?

CommonLib/Error/GREYError+Private.h
```diff
@@ -60,9 +60,9 @@ GREY_EXTERN NSString *const kErrorFunctionNameKey;
 GREY_EXTERN NSString *const kErrorUserInfoKey;
 
 /**
- * Key used to retrieve the bundle ID from an error object dictionary.
+ * Key used to retrieve the error info dictionary from an error object dictionary.
  */
-GREY_EXTERN NSString *const kErrorBundleIDKey;
+GREY_EXTERN NSString *const kErrorErrorInfoKey;
```
tirodkar, line 65: Is this key used anywhere? I couldn't find it.

GREYErrorFormatter.m
tirodkar: You would not be creating a new array.

GREYErrorFormatter.m
tirodkar: They can do it when repeatedly performing an interaction and checking the error description. This is only called internally in this function - this can be made a static function as well.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,259 @@
...
+  }
+  return [GREYObjectFormatter formatDictionary:[_error grey_descriptionDictionary]
+                                        indent:kGREYObjectFormatIndent
+                                     hideEmpty:YES
+                                      keyOrder:nil];
+}
+
+#pragma mark - Private Methods
+
+- (NSSet<NSString *> *)loggerKeys {
```
tirodkar, line 85: Comment.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,259 @@
...
+
+// Example: errorLogEntryWithNewlineBefore:NO entry:@"1" colon:YES space:YES
+// secondNewLine:YES secondEntry:@"2" -> @"1: \n2\n"
+// Notice it always adds a \n to the end, so there is no need to supply that as a parameter.
+- (NSString *)errorLogEntryWithNewlineBefore:(BOOL)lineBefore
+                                       entry:(NSString *)entry
+                                       colon:(BOOL)colon
+                                       space:(BOOL)space
+                               secondNewLine:(BOOL)secondNewLine
+                                 secondEntry:(NSString *)secondEntry {
```
tirodkar, line 203: Won't this be expanded ad nauseam in case newer keys are to be added? I think it would be better to have the NSString contain the spacing and newline info instead.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,259 @@
...
+    [log appendString: @":"];
+  }
+  if (space) {
+    [log appendString: @" "];
+  }
+  if (secondNewLine) {
+    [log appendString: @"\n"];
+  }
+  if (secondEntry) {
+    [log appendString: secondEntry];
```
tirodkar, line 221: rm

TestLib/Exception/GREYFailureFormatter.h
tirodkar:
```
 * @param testCase       The test case where the failure happens.
 * @param failureLabel   The label to be associated with the error.
 * @param failureName    The name to be associated with the error.
 * @param functionName   The name of the function where the error is generated.
 * @param filePath       The path to the source code file where the error happens.
 * @param lineNumber     The line number of the source code file when the error happens.
 * @param stackTrace     The stack trace of the executable when the error happens.
 * @param appScreenshots The dictionary of app screenshot names and image paths
 *                       when the error happens.
 * @param errorDescription         Extra message to be included in the message.
```

The formatting seems wonky in the raw file as well.

TestLib/Exception/GREYFailureFormatter.m
```diff
@@ -23,24 +23,6 @@
 
 @implementation GREYFailureFormatter
 
-+ (NSString *)formatFailureForError:(GREYError *)error
```
tirodkar, line 26: Can you send a separate PR for this?

CommonLib/Error/GREYError+Private.h
willsaid-google: It is - see GREYError.m:128 and GREYElementInteraction.m:818 for example

willsaid-google: kErrorBundleIDKey was not being used so i removed it.

TestLib/Exception/GREYFailureFormatter.m
willsaid-google: I had to change the method parameter to not be variadic in order for FailureFormatter to work with the rest of the changes. And when i changed that parameter (to not be variadic), it was then the exact same method as the other method, so i had to remove it. So, I think this removal should be in this PR.

TestLib/Exception/GREYFailureFormatter.h
willsaid-google: Fixed

GREYErrorFormatter.m
willsaid-google: That would be indeed fancier, but what are the benefits? In Swift, an enum would make more sense because i could specify the enum type as the array generic element type. In Obj C, no build-time safety is enforced for the integer enum approach over the static string constant approach, right?

AlbertWang0116: It's mainly for the efficiency. To verify existence from a NSSet, you need to call string's `hash`, which is O(len(string)); OTOH bit operation is just O(1).

Is there any difference on build-time safety here? I take for either const string or enum integer, you always use it internally. Can you elaborate more?

GREYErrorFormatter.m
```diff
@@ -0,0 +1,145 @@
...
+  if ([keys containsObject:kErrorFormatterElementMatcherKey]) {
+    NSString *elementMatcher = _error.userInfo[kErrorDetailElementMatcherKey];
+    if (elementMatcher) {
+      [logger addObject:[NSString stringWithFormat:@"%@:\n%@", kErrorDetailElementMatcherKey,
+                         elementMatcher]];
+    }
+  }
+  
+  if ([keys containsObject:kErrorFormatterAssertionCriteriaKey]) {
+    NSString *assertionCriteria = _error.userInfo[kErrorFormatterAssertionCriteriaKey];
```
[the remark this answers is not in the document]
willsaid-google: Resolved. Thanks

TestLib/Exception/GREYDefaultFailureHandler.m
```diff
@@ -81,24 +132,48 @@ - (void)handleException:(GREYFrameworkException *)exception details:(NSString *)
...
-                                                    functionName:nil
-                                                      stackTrace:stackTrace
-                                                  appScreenshots:screenshotPaths
-                                                       hierarchy:appUIHierarchy
-                                                          format:@"%@", logMessage];
+- (void)handleException:(GREYFrameworkException *)exception details:(NSString *)details {
+  GREYThrowOnNilParameter(exception);
+  id currentTestCase = [XCTestCase grey_currentTestCase];
+  GREYFailureScreenshots *screenshotPaths = [self screenshotPathsForException:exception];
+  NSArray<NSString *> *stackTrace = [NSThread callStackSymbols];
```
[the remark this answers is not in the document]
willsaid-google: Do you have a link to where we deleted this?

GREYErrorFormatter.m
willsaid-google: In Swift it'd make more sense to use an Enum because i could return an array of that enum type, but in Obj C, that is not possible. Good point on the performance difference. @tirodkar @chan-park what do you think?

willsaid-google: Pushed the changes.

GREYErrorFormatter.m
willsaid-google: Fixed

GREYErrorFormatter.m
willsaid-google: Fixed

GREYErrorFormatter.m
willsaid-google: Look at the most recent commit. I think it fixes the issue.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,178 @@
...
+ * Keys used by GREYErrorFormatter to format a GREYError's userInfo properties.
+ * These states are not mutually exclusive and can be combined together using Bitwise-OR to
+ * represent multiple states.
+ * If more than 32 options exists, change the bitshifted values to UL.
+ */
+typedef NS_OPTIONS(NSUInteger, GREYErrorFormatterKeys) {
+  GREYErrorFormatterNone = 0,
+  /**
+   * Exception Reason
+   */
```
AlbertWang0116, line 38: You can use /** ... */ for short doc string.

GREYErrorFormatter.m
```diff
@@ -0,0 +1,178 @@
...
+- (NSString *)loggerDescriptionForKeys:(NSUInteger)keys {
+  NSMutableArray<NSString *> *logger = [[NSMutableArray alloc] init];
+  
+  if (keys & GREYErrorFormatterExceptionReasonKey) {
+    NSString *exceptionReason = _error.localizedDescription;
+    [logger addObject:[NSString stringWithFormat:@"\n%@", exceptionReason]];
+  }
+  
+  if (keys & GREYErrorFormatterRecoverySuggestionKey) {
+    NSString *recoverySuggestion = _error.userInfo[kErrorDetailRecoverySuggestionKey];
```
chan-park, line 122: Why are we using different key to fetch the value from `userInfo`?

CommonLib/Error/GREYError+Private.h
tirodkar, line 65: Where is this used? errorInfo can be obtained without this change. Please revert if not needed.

CommonLib/Error/GREYError.m
```diff
@@ -108,10 +109,7 @@ + (instancetype)errorWithDomain:(NSString *)domain
 }
 
 - (NSString *)description {
```
tirodkar, line 111: A user can also check for error.localizedDescription. If you NSLog this, it will still print the contents of the errorInfo dictionary. You should also override localizedDescription and print the new GREYError description you are printing, else call `[super localizedDescription];`.

CommonLib/Error/GREYErrorFormatter.m
```diff
@@ -0,0 +1,197 @@
...
+  return [reason containsString:@"the desired element was not found"];
+}
+
+#pragma mark - Static Functions
+
+// The keys whose values should be supplied in the formatted error output.
+static NSUInteger loggerKeys(GREYError *error) {
+  if ([error.domain isEqualToString:kGREYInteractionErrorDomain] &&
+       error.code == kGREYInteractionElementNotFoundErrorCode) {
+     return GREYErrorFormatterFlagExceptionReason |
```
tirodkar, line 98: In the other PR's this key for Exception Reason get repeated. Are all the keys below always populated irrespective of the error type? If not, in the LoggerDescription method below - can you print a log statement based on if its source is nil / not nil?

e.g.

```
if (keys & GREYErrorFormatterFlagExceptionReason) {
    NSString *exceptionReason = error.localizedDescription;
    [logger addObject:[NSString stringWithFormat:@"\n%@", exceptionReason]];
  }
```
Can this instead be - 

```
NSString *exceptionReason = error.localizedDescription;
if (exceptionReason) {
    [logger addObject:[NSString stringWithFormat:@"\n%@", exceptionReason]];
  }
```

CommonLib/Error/GREYErrorFormatter.m
```diff
@@ -0,0 +1,197 @@
...
+            GREYErrorFormatterFlagElementMatcher |
+            GREYErrorFormatterFlagCriteria |
+            GREYErrorFormatterFlagSearchActionInfo |
+            GREYErrorFormatterFlagUnderlyingError |
+            GREYErrorFormatterFlagUIHierarchy;
+  }
+  GREYFatalAssertWithMessage(false, @"Error Domain and Code Not Yet Supported");
+}
+
+static NSString *formattedHierarchy(NSString *hierarchy) {
```
tirodkar, line 109: http://google.github.io/styleguide/objcguide.html#function-names

```
Function names should start with a capital letter and have a capital letter for each new word (a.k.a. “camel case” or “Pascal case”).
```

TestLib/EarlGreyImpl/GREYElementInteractionErrorHandler.m
```diff
@@ -78,7 +79,13 @@ void GREYHandleInteractionError(__strong GREYError *interactionError,
       id<GREYFailureHandler> failureHandler =
           [NSThread mainThread].threadDictionary[GREYFailureHandlerKey];
       // TODO(b/147072566): Will show up a (null) in rotation.
-      [failureHandler handleException:exception details:matcherDetails];
+      if (GREYShouldUseErrorFormatterForError(interactionError)) {
+        // Eventually this check will not be needed, and the error's description will
+        // always be passed as the `details`
+        [failureHandler handleException:exception details:interactionError.description];
+      } else {
```
tirodkar, line 86: Since interactionError.description is being used directly you can move this to line 54.

     if (GREYShouldUseErrorFormatterForError(interactionError)) {
        [failureHandler handleException:exception details:interactionError.description];
      } else {
        NSMutableString *matcherDetails;
        ......
      }

CommonLib/Error/GREYError.m
willsaid-google: Shouldn't this go into a separate PR?

tirodkar: Add a TODO for it. This shouldn't impact calling localizedDescription as of now that users are doing.

TestLib/EarlGreyImpl/GREYElementInteractionErrorHandler.m
willsaid-google: Resolved. It had to be more complex than just moving it to 54, so take a look

CommonLib/Error/GREYErrorFormatter.m
willsaid-google: Resolved, PTAL

CommonLib/Error/GREYErrorFormatter.m
willsaid-google: Good point, resolved. Also resolved for underlying error, since both of those will always be present.

CommonLib/Error/GREYError+Private.h
willsaid-google: Only inside GREYError.m. Removed it.

CommonLib/Error/GREYError.m
willsaid-google: Resolved

willsaid-google:
Note that builds are failing due to an unrelated bug in the earlgrey2 branch.

willsaid-google:
Fixed merge conflicts with eg2

CommonLib/Error/GREYError+Private.h
```diff
@@ -59,11 +59,6 @@ GREY_EXTERN NSString *const kErrorFunctionNameKey;
  */
 GREY_EXTERN NSString *const kErrorUserInfoKey;
 
-/**
- * Key used to retrieve the errorInfo dictionary from an error object dictionary.
- */
-GREY_EXTERN NSString *const kErrorErrorInfoKey;
```
tirodkar, line 65: Sync this.

TestLib/Exception/GREYDefaultFailureHandler.m
```diff
@@ -81,23 +94,46 @@ - (void)handleException:(GREYFrameworkException *)exception details:(NSString *)
...
+}
 
-  NSString *appUIHierarchy = [exception.userInfo valueForKey:kErrorDetailAppUIHierarchyKey];
-  // For calls from GREYAsserts in the test side, the hierarchy must be populated here.
-  if (!appUIHierarchy) {
-    appUIHierarchy = [GREYElementHierarchy hierarchyString];
+- (NSString *)consoleLogForException:(GREYFrameworkException *)exception
+                             details:(NSString *)details
+                     screenshotPaths:(GREYFailureScreenshots *)screenshotPaths
+                      appUIhierarchy:(NSString *)appUIHierarchy
```
tirodkar, line 103: You don't need to pass the appUIHierarchy and screenshots here as they are computed only in this method. You can move the calls to 

```
GREYFailureScreenshots *screenshotPaths = [self screenshotPathsForException:exception];
NSString *appUIHierarchy = [self appUIHierarchyForException:exception];
```

Inside this function where there are being used. So, `[self appUIHierarchyForException:exception];` goes where the formatFailure... method is called.

TestLib/Exception/GREYDefaultFailureHandler.m
```diff
@@ -81,23 +94,46 @@ - (void)handleException:(GREYFrameworkException *)exception details:(NSString *)
...
                                                          toDirectory:screenshotDir];
   NSAssert(screenshotPaths, @"Screenshots must be present");
+  return screenshotPaths;
+}
 
-  NSString *appUIHierarchy = [exception.userInfo valueForKey:kErrorDetailAppUIHierarchyKey];
-  // For calls from GREYAsserts in the test side, the hierarchy must be populated here.
-  if (!appUIHierarchy) {
-    appUIHierarchy = [GREYElementHierarchy hierarchyString];
+- (NSString *)consoleLogForException:(GREYFrameworkException *)exception
```
tirodkar, line 100: Please add comments for this and other functions created.

TestLib/Exception/GREYDefaultFailureHandler.m
```diff
@@ -64,7 +64,20 @@ - (void)handleException:(GREYFrameworkException *)exception details:(NSString *)
     NSString *logDetails = [NSString stringWithFormat:@"Exception Details: %@", details];
     [logMessage appendString:logDetails];
   }
+  
+  return logMessage;
+}
 
+- (NSString *)appUIHierarchyForException:(GREYFrameworkException *)exception {
```
tirodkar, line 71: Comment this and the below method.

CommonLib/Error/GREYErrorFormatter.m
```diff
@@ -0,0 +1,141 @@
...
+}
+
+#pragma mark - Public Functions
+
+BOOL GREYShouldUseErrorFormatterForError(GREYError *error) {
+  return [error.domain isEqualToString:kGREYInteractionErrorDomain] &&
+          error.code == kGREYInteractionElementNotFoundErrorCode;
+}
+
+BOOL GREYShouldUseErrorFormatterForExceptionReason(NSString *reason) {
```
tirodkar, line 61: Looking at https://github.com/willsaid-google/EarlGrey/pull/6/files, this function will definitely become bloated. Instead, we could add a prefix to this formatter hierarchy as we roll it out piecemeal per error.

So, in LoggerDescription - you can start the string with:

"GREYErrorFormatter Information:\n" and trim the prefix in the failure handler as a temporary hack. It will remove the need to add more strings in here.

willsaid-google: Trimming the prefix isn't reliable since, for example, a user could call GREYError's description. At what point would we know when to trim?...Only when the default failure handler is used. So, i propose we use a flag that is actually useful, and won't need to be trimmed, such as "EarlGrey Encountered an Error:"

willsaid-google: Pushed. Please check out the changes

TestLib/Exception/GREYDefaultFailureHandler.m
willsaid-google: done

TestLib/Exception/GREYDefaultFailureHandler.m
willsaid-google: done

TestLib/Exception/GREYDefaultFailureHandler.m
willsaid-google: Resolved

CommonLib/Error/GREYError+Private.h
willsaid-google: Resolved

tirodkar:
This is good for merging internally. We'll break this us - ErrorFormatter first and then the failure handlers. 

There are small formatting / comment fixes which we can do from our side, so you might have very slight conflicts. Otherwise, we can now continue with further error formatter changes.

tirodkar:
This is now merged internally 😃. Closing.

Issue #1357, opened 2020-06-23
Internal private APIs are now marked as "direct" in iOS 14


LeoNatan:
Meaning that most synchronization points, where Earl Grey (and other projects) swizzles so it can monitor the system state, are no longer available for discovery using the ObjC runtime system (reflection), and can no longer be swizzled because the underlying functions are statically called inside the system.

I wanted to start this discussion as early as now, because this will create major issues for this project and should be looked at now.

tirodkar:
Thanks a lot for filing this @LeoNatan. I'm doing an audit to find out what methods are no longer supported for Swizzling. Till now I've only had reports of some UITouch methods failing.

patchthecode:
Do you know or have the issue# of the touch methods failing?
I am assuming it was referenced in some other issue.

steipete:
I'm doing some investigation here; it seems `UITouch` has been changed quite a bit.

KIF is facing the same issue: https://github.com/kif-framework/KIF/issues/1152

Just commenting out `setIsTap:` and `_setIsFirstTouchForView:` doesn't fix the issue, taps no longer work.

steipete:
This seems to be an okay replacement - need more testing tho: (See the PSPDFKit marked part)

```objc
@implementation UITouch (GREYAdditions)

- (id)initAtPoint:(CGPoint)point relativeToWindow:(UIWindow *)window {
  GREYThrowOnNilParameter(window);

  point = CGPointAfterRemovingFractionalPixels(point);

  self = [super init];
  if (self) {
    [self setTapCount:1];
    [self setPhase:UITouchPhaseBegan];

    [self setWindow:window];
    [self _setLocationInWindow:point resetPrevious:YES];
    [self setView:[window hitTest:point withEvent:nil]];
    [self setTimestamp:[[NSProcessInfo processInfo] systemUptime]];

    // PSPDFKit: This was changed in iOS 14.
    if ([self respondsToSelector:@selector(_setIsFirstTouchForView:)]) {
        [self setIsTap:YES];
        [self _setIsFirstTouchForView:YES];
    } else {
        [self _setIsTapToClick:YES];

        // We modify the touchFlags ivar struct directly.
        // First entry is _firstTouchForView
        Ivar flagsIvar = class_getInstanceVariable(object_getClass(self), "_touchFlags");
        ptrdiff_t touchFlagsOffset = ivar_getOffset(flagsIvar);
        char *flags = (__bridge void *)self + touchFlagsOffset;
[...]

PR #1431, opened 2020-09-01
Backport iOS 14 fixes to EarlGrey v1

xavierjurado:
This PR backports all the issues described in https://github.com/google/EarlGrey/issues/1357 to EarlGrey v1. Where possible I've cherry-picked the original commit:

* XCTestCase::recordIssue support: https://github.com/google/EarlGrey/pull/1377
* UITouch private API changes (based on @steipe approach published in https://github.com/google/EarlGrey/issues/1357)
* Disable caret blink animation: https://github.com/google/EarlGrey/pull/1404/

Additionally I've also backported/fixed other small issues:

* Duplicate TextField id's: https://github.com/google/EarlGrey/pull/1391/ 
* Remove duplicated/invalid file references from Build Phase
* Explicitly remove support for arm64 simulator (Apple Silicon) (not supported by ochamcrest)
* Configure indentation to match existing guidelines
* Fix deprecated developmentRegion
* Fix weird documentation error due to an escaped character

Let me know if you want to add/remove/split anything.

(I've signed the CLA)


joshuapoq:
Thank you for this! 
I was able to get our repos working with this for a few hours earlier but after clearing all cache to match CIs failures all accessibility frames appear to be `.zero` so all `sufficientlyVisible()` and `interaction()` are failing.
Has anyone else observed this?

xavierjurado:
We experienced the same issue locally, as far as I can tell is due to some ¿bug? on the iOS simulator. I can't seem to find the post where I read about it, but we managed to fix the issue either by restarting the simulators or (don't laugh)  launching the accessibility inspector and inspecting a view; from that point onwards everything worked flawlessly (and whatever happened persisted through executions). 

Luckily for us we didn't experienced the same issue on our CI (Travis Enterprise), so not sure how to proceed 🤔 .

joshuapoq:
Honestly I've been wracking my brain what's changed and was seriously suspecting the simulator. Haha you're comments definitely put my frustrations at ease. 😂 Am not at all surprised that booting up the inspector might have fixed this.
Thanks for your quick reply @xavierjurado! Will try a few more things.

xavierjurado:
Hahaha, glad it worked! It would be cool to know what on earth is happing behind the scenes, chances are it's something we can fix setting some user defaults.

tirodkar:
@xavierjurado Is your CI environment the exact same as your local one where you booted the accessibility inspector. EarlGrey needs the accessibility inspector so as to read the actual ax frames / info of the elements. I'm afraid there might be a bigger breakage going on with EarlGrey 1.0 and the ax inspector loading here.

https://github.com/google/EarlGrey/blob/master/EarlGrey/Core/GREYAutomationSetup.m#L155 The code here should crash in case the inspector isn't launched.

joshuapoq:
I found that code (I assume it has a bug, but haven't investigated) and also took a look at KIF who have a similar thread explaining the issue:
https://github.com/kif-framework/KIF/issues/944

Anyway I've fixed loading the accessibility within our project with the following code in our base test case - I know I should probably use the principal class in the info plist but this does the trick and only runs once:
```swift
class EGTestCase: XCTestCase {
    static let fixAccessibility: Void = {
        let libAccessibilityPath = NSString("/usr/lib/libAccessibility.dylib").fileSystemRepresentation
        guard let handle = dlopen(libAccessibilityPath, RTLD_LOCAL) else { return }
        
        typealias AXSSetAutomationEnabled = @convention(c) (Bool) -> Void
        guard let _AXSSetAutomationEnabled = dlsym(handle, "_AXSSetAutomationEnabled") else { return }
        let setAutomationEnabled = unsafeBitCast(_AXSSetAutomationEnabled, to: AXSSetAutomationEnabled.self)
        
        setAutomationEnabled(true)
    }()
    
    open override func invokeTest() {
        Self.fixAccessibility
        super.invokeTest()
    }
```

tirodkar:
Thanks. I'm not sure if that will work on devices though. There have been failures previously on device with this code.

joshuapoq:
Good point @tirodkar, thanks for pointing that out! Our projects don't run these tests on devices at the moment but we're looking to move to UI tests so will mark this in the code!

tirodkar:
EarlGrey 2.0 btw does not have this issue as XCUITest provides accessibility for us.

LeoNatan:
Wasn't there a way to trigger the accessibility system load from some `UIApplication` API?

Perhaps https://github.com/LeoNatan/Apple-Runtime-Headers/blob/5e50ad05dfd7d7b69fc2e0e685765fc054166b3c/iOS/PrivateFrameworks/UIKitCore.framework/UIApplication.h#L838 ?

xavierjurado:
The issue turned out to be much simpler: I accidentaly removed the invocation to `[[GREYAutomationSetup sharedInstance] prepareOnLoad];` when cherry-picking c9517d5f55f1cd1ec67c6522e6be214fa5dbdc93, which in turn prevented `grey_enableAccessibilityForSimulator` from being ever invoked. Fixed in 2e28af9.

xavierjurado:
@rpitting Although EarlGrey is open source it is distributed as a binary framework via Cocoapods. That means that even if you pointed to our fork you'd still be downloading Google's official precompiled version unless we changed the podspec to point to our own binary. This is something we have not done yet because we'd like to merge these changes here instead.

In the meantime what we have done and what I suggest you is to compile the framework manually and use it instead of the Cocoapods version (ie: manual integration). I'll let you know if/when we distribute our own binary!

ravimandala:
We are looking forward to have this PR merged and a subsequent release that can be used on Xcode 12. Thanks @xavierjurado for working on this.

tirodkar:
Folks, we are not planning to do another EarlGrey 1.0 CocoaPods release. The code has not been maintained in over a year and a half.

xavierjurado:
I feel like it would be a good idea to put a deprecation statement somewhere (in the README or the issue/PR template maybe?) so that people are aware of this and do not accidentally integrate with EarlGrey v1. Adding the `spec.deprecated = true` could also be useful. Thanks for the clarification @tirodkar!

Issue #1510, opened 2020-11-24
EarlGrey 2.2.0 | Cocoapods | Can't build UI Tests target when use use_frameworks! flag in Podfile


kwlx:
After updating EarlGrey to 2.2.0, I cannot build a UI Tests target. This build failed with an error:

> Include of non-modular header inside framework module 'EarlGreyTest.GREYHostApplicationDistantObject': '/xxx/Library/Developer/Xcode/DerivedData/XXX-fwetoamgsahlylbaeymzuyfiookw/Build/Products/Debug-iphonesimulator/eDistantObject/eDistantObject.framework/Headers/EDOClientService.h'

This only happens when use_frameworks! flag is used in Podfile in UI Tests target. Unfortunately I have to use this flag because of a different library.

It can be reproduce in EarlGreyExample from earlgrey2 branch.
<img width="673" alt="Zrzut ekranu 2020-11-24 o 08 30 02" src="https://user-images.githubusercontent.com/41103151/100062327-ab776000-2e2f-11eb-870c-d15a3fbfe4d4.png">

Xcode 12.2
Cocoapods 1.10.0

tirodkar:
You only need `EarlGreyTest` in the test component of your PodFile. Can you not remove use_frameworks from it? My guess is that you have a global Podfile config from where use_frameworks is added. Do you need the `different library` in your tests?

kwlx:
Yes, unfortunately I need a different library in my tests and it needs the use_framework flag.

PR #1515, opened 2020-11-27
Fix import eDistantObject for cocoapods

konradzdunczyk:
Fix for #1510

### DIFF
--- a/CommonLib/DistantObject/GREYHostApplicationDistantObject.h
+++ b/CommonLib/DistantObject/GREYHostApplicationDistantObject.h
@@ -16,11 +16,11 @@
 
 #import <Foundation/Foundation.h>
 
-// Use modular imports for CocoaPods release.
-// #import <eDistantObject/EDOClientService.h>
-// End CocoaPods modular imports
-
+#if COCOAPODS
+#import <eDistantObject/EDOClientService.h>
+#else
 #import "EDOClientService.h"
+#endif // COCOAPODS
 
 @class EDOHostService;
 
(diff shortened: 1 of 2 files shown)


CommonLib/DistantObject/GREYHostApplicationDistantObject.h
```diff
@@ -16,11 +16,11 @@
 
 #import <Foundation/Foundation.h>
 
-// Use modular imports for CocoaPods release.
-// #import <eDistantObject/EDOClientService.h>
-// End CocoaPods modular imports
-
+#if COCOAPODS
```
tirodkar, line 19: Is this a define CocoaPods provides us?

konradzdunczyk: Yes

image

shaneong:
Please merge this PR as it is blocking our use case where we have a separate Pod Framework in our Podfile hence EarlGrey is failing to build without this fix at the moment.

shaneong:
How long does it take for changes to get merged in to CocoaPods? Am still not observing this change in installed EarlGreyTest Pod

tirodkar:
@shaneong we merged this internally so closing this. We'll have to do another CocoaPods release which I can check with the team on the duration of.

shaneong:
Can you prioritize this fix as being able to use EarlGrey in conjunction with use_frameworks! should be a common ask in CocoaPods, or at least provide a workaround solution in the meantime?

PR #1735, opened 2022-03-31
Fix wrong path in Scripts/download_deps.sh

623637646:
Fix the wrong path. Then `EarlGrey.xcodeproj` works properly.


tirodkar:
Where are you running this file from? This has to be called from the root directory of the EarlGrey folder.

623637646:
I see. I called it in the `Scripts` folder.  Thanks.

PR #1973, opened 2023-09-06
Tag `-selectElementWithMatcher:` with `NS_WARN_UNUSED_RESULT`

sigito:
We often run into a human error, when an element is selected by not asserted on or acted upon. With the unused result attribute, the person writing a test would get a nag that no action was performed on the element.


tirodkar:
Will port to eg2 as well.

PR #2117, opened 2025-08-20
Define rotateDeviceToOrientation using rotateInterfaceToOrientation

### LINKED ISSUE #2111: Inconsistency in `-[EarlGreyImpl rotateDeviceToOrientation: error:]`
The documentation and the implementation differs.
According to the [documentation](https://github.com/google/EarlGrey/blob/d5b5f18f3cf231f3b3c53a898780926aaeed2905/TestLib/EarlGreyImpl/EarlGreyImpl.h#L274), any orientation is a valid input, except for unknown. 

Except that the orientations face up/down are not actually valid. They cause an error.

Indeed, those two Interface orientations are mapped to the unknown devices orientations. And so the behavior in this case is exactly the same as if the user used unknown input.

So either the implementation or the documentation should be changed.

I’m personally more in favor of doing the following change:
* the current method, taking the device orientation as input, would have its documentation changed to note that face up/down are also not authorized input. 
* and potentially, split this function in two. One that takes the interface orientation as input and the other one that takes the device orientation as input. The latter would be implemented using the former, and marked as deprecated. The latter would indeed accept any value except for unknown.

An alternative solution would be to state that there is some default interface orientation associated with face up/down. But I don’t think that would be intuitive for the user so I’m slightly against it. 

This is related to Chromium bug crbug.com/438558888

Arthur-Milchior:
This ensures that the deprecated method is not called anymore. This also avoids to transform the orientation from interface to device to interface.

Fixed #2111

### DIFF
--- a/TestLib/EarlGreyImpl/EarlGreyImpl.m
+++ b/TestLib/EarlGreyImpl/EarlGreyImpl.m
@@ -265,22 +265,21 @@ static BOOL ExecuteSyncBlockInBackgroundQueue(BOOL (^block)(void)) {
 
 #if TARGET_OS_IOS
 
-- (BOOL)rotateInterfaceToOrientation:(UIInterfaceOrientation)interfaceOrientation
-                               error:(NSError **)error {
-  UIDeviceOrientation deviceOrientation =
-      [GREYConstants deviceOrientationForInterfaceOrientation:interfaceOrientation];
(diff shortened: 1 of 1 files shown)


Arthur-Milchior:
Closing as this was essentially merged in edd035fa9ca15fa06c425af849acd9dfae6159c3

